PR #6, opened 2020-07-02
Configure black to comply with flake8

EQt:


johanneskoester:
The tests fail because someone committed the test files in the wrong place. This is fixed in PR #7.

PR #7, opened 2020-07-02
Separate annotation filtering and test framework updates.

johanneskoester:
* Implement separate expression for annotation filtering. 
* Make both filter expressions optional. 
* Adapt testing framework and fix already committed test.


vembrane/__init__.py
```diff
@@ -23,45 +23,93 @@
...
+    return []
+
+
+def parse_annotation_entry(entry: str):
+    return list(map(str.strip, entry.split("|")))
+
+
+def filter_annotation_entries(entries: list, ann_filter_expression: str):
+    for entry in entries:
+        env = dict(zip(annotation_keys, parse_annotation_entry(entry)))
```
tedil, line 39: `annotation_keys` is not defined here; either move this def into the filter_vcf definition or make annotation_keys a param (or a global...)

johanneskoester: good catch.

PR #10, opened 2020-07-02
make INFO-level environment available in ann-filtering.

johanneskoester:
Move ann entries below an ANN key.


tedil:
To make the linter happy, the typing.Dict import needs to go ^^

PR #11, opened 2020-07-02
One expression

christopher-schroeder:
please review


vembrane/__init__.py
```diff
@@ -58,44 +44,30 @@ def filter_vcf(
...
-                        )
-                    )
-                    if not filtered_ann:
-                        # skip this record if filter removed all annotations
-                        continue
-                    record.info["ANN"] = filtered_ann
-
-            yield record
+            if eval(expression, globals_whitelist, env):
+                filtered_ann.append(annotation)
```
tedil, line 64: perhaps we should keep a list/bitvector of annotations to remove instead

vembrane/__init__.py
```diff
@@ -58,44 +44,30 @@ def filter_vcf(
...
-        ann = dict(
-            zip(
-                annotation_keys,
-                zip(
-                    *(
-                        parse_annotation_entry(entry)
-                        for entry in record.info.get("ANN", [])
-                    )
-                ),
+        # setup filter expression env
```
johanneskoester, line 47: We should keep building the env dict here with the global entries. This does not need to be redone in each iteration below. Below, we can just overwrite the ANN key.

PR #17, opened 2020-07-02
argparse parameter for keyname

christopher-schroeder:
annotations = record.info.get(ann_key, [])
crashes if ann_key does not exist

solution:
annotations = dict(record.info).get(ann_key, [])

additional: parameter for annotation key name

### DIFF
--- a/vembrane/__init__.py
+++ b/vembrane/__init__.py
@@ -113,7 +113,7 @@ def main():
         help="Output format.",
     )
     parser.add_argument(
-        "--ann_key",
+        "--annotation-key",
         "-k",
         metavar="FIELDNAME",
         default="ANN",
--- a/vembrane/__init__.py
+++ b/vembrane/__init__.py
@@ -48,7 +48,9 @@ def eval_expression(
         return False
 
 
-def filter_vcf(vcf: VariantFile, expression: str, ann_key:str="ANN") -> Iterator[VariantRecord]:
+def filter_vcf(
+    vcf: VariantFile, expression: str, ann_key: str = "ANN"
+) -> Iterator[VariantRecord]:
     header = vcf.header
 
     env = dict()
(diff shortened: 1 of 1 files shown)


vembrane/__init__.py
```diff
@@ -110,6 +112,14 @@ def main():
         choices=["vcf", "bcf", "uncompressed-bcf"],
         help="Output format.",
     )
+    parser.add_argument(
+        "--ann_key",
```
tedil, line 116: Should be `--ann-key` (or even `--annotation-key` since it's a long option anyway), i.e. hyphen instead of underscore.

vembrane/__init__.py
```diff
@@ -65,10 +67,10 @@ def filter_vcf(vcf: VariantFile, expression: str) -> Iterator[VariantRecord]:
         env["POS"] = record.pos
         env["REF"], env["ALT"] = chain(record.alleles)
         for key in record.info:
-            if key != "ANN":
+            if key != ann_key:
                 env[key] = record.info[key]
 
-        annotations = record.info.get("ANN", [])
+        annotations = dict(record.info).get(ann_key, [])
```
tedil, line 73: Why does it crash? Looking at the [`VariantRecordInfo.get`](https://github.com/pysam-developers/pysam/blob/2282e99dde733360491cedfbfa6a8b6a4d191d6d/pysam/libcbcf.pyx#L2657) method, it should work without the dict wrapping

christopher-schroeder: it should, but it doesn't, give it a try

tedil: on that note, make sure to add a test for that

vembrane/__init__.py
```diff
@@ -80,7 +82,7 @@ def filter_vcf(vcf: VariantFile, expression: str) -> Iterator[VariantRecord]:
             continue
```
tedil, line 82: If `annotations == []` (which happens if there's no ANN INFO field in the header), the record will never get filtered, even if the expression evaluates to false

PR #18, opened 2020-07-02
parameter to keep unmatched annotations

christopher-schroeder:
this parameter let annotations unchanged and only removes variants, when no annotation passes the expression
when "--keep-unmatched" is enabled, vembrane now behaves like filter_vep (where you can remove annotations by its "only_matched" flag).

### DIFF
--- a/vembrane/__init__.py
+++ b/vembrane/__init__.py
@@ -48,7 +48,7 @@ def eval_expression(
         return False
 
 
-def filter_vcf(vcf: VariantFile, expression: str) -> Iterator[VariantRecord]:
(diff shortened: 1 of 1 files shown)


vembrane/__init__.py
```diff
@@ -23,32 +30,40 @@
 }
 
 
-def get_annotation_keys(header: VariantHeader) -> List[str]:
+def get_annotation_keys(header: VariantHeader,) -> List[str]:
```
tedil, line 33: did black introduce the trailing comma? i.e. was there a style change?

christopher-schroeder: clickbait: the linter hates THESE commas

PR #20, opened 2020-07-03
statistics

christopher-schroeder:
I implemented a --statistics [n] parameter, which takes the first n reads and counts the occurrence of the different values after filtering instead of writing the vcf. This is especially useful, when you want to check, if you have missed something in your expression.
Now that I think about it, it should additionally write the statistics in a separate file, instead of just printing it and not writing any filtered vcf. But this is open for discussion.

Please also note that --statistics 0 is very useful to show the annotation keys!


vembrane/__init__.py
```diff
@@ -29,7 +31,7 @@ def get_annotation_keys(header: VariantHeader) -> List[str]:
             return list(map(str.strip, rec.get("Description").split("'")[1].split("|")))
     return []
 
-
+@lru_cache
```
tedil, line 34: Good idea! Should have been in a different PR, but whatever.
My only concerns are:
 - Default bound for the cache (how many entries are cached til old ones get ditched?)
 - Did you check if there even are identical annotation strings "in the wild", i.e. if this actually caches anything useful?

vembrane/__init__.py
```diff
@@ -84,6 +86,30 @@ def filter_vcf(vcf: VariantFile, expression: str) -> Iterator[VariantRecord]:
         yield record
 
 
+def statistics(vcf, records, n):
```
tedil, line 89: I'd personally have called it `stats` instead

vembrane/__init__.py
```diff
@@ -84,6 +86,30 @@ def filter_vcf(vcf: VariantFile, expression: str) -> Iterator[VariantRecord]:
...
+        if i >= n: # process only the first n records
+            break
+        for annotation in record.info["ANN"]:
+            for key, value in zip(annotation_keys, parse_annotation_entry(annotation)):
+                if value:
+                    counter[key][value] += 1
+    
+    for key, subdict in counter.items():
+        print(key)
+        if len(subdict) <= 10:
```
tedil, line 106: Is 10 a good threshold?
Apart from that, I prefer introducing constants (somewhere), makes it easier to change thresholds such as this one later on.

vembrane/__init__.py
christopher-schroeder: default is 128, i thought it would be enough. We could also reduce it to 32.
i introduced the lru cache, because statistics parse_annotation_entry is called twice now. In "statistics" and "filter_vcf". In theory we could manage this in a different way and somehow pass the split records to statistics, but I think they way is is currently implemented, is the most pythonic way.

vembrane/__init__.py
christopher-schroeder: I prefer to make this adjustable by the user :)

christopher-schroeder: I will change the branch, so that it will write an additional statistic file.

vembrane/__init__.py
tedil: ...but that doesn't really matter

vembrane/__init__.py
tedil: > I prefer to make this adjustable by the user :)

as long as the cli doesn't have too many options / isn't cluttered ;)

christopher-schroeder: let me show you the snakemake --help
:D

vembrane/__init__.py
```diff
@@ -33,6 +37,7 @@ def get_annotation_keys(header: VariantHeader,) -> List[str]:
     return []
 
 
+@lru_cache(maxsize=32)
```
tedil, line 40: as I said, we'll have to run this on actual (large) vcf files to find a good value

PR #23, opened 2020-07-03
reraise errors

tedil:


christopher-schroeder:
I would say that Johannes request is a feature. We should implement as many understandable error messages as possible for all kind of DAUs

tedil:
> So if we anyway exit with an error here, I think we should also print a proper error message and not a stack trace. Maybe we should create a new error type, with a proper message, and catch that in the main function, only print the error and exit with exit code 1.

I introduced some new error types, only print the error and exit with code 1 ;)

PR #24, opened 2020-07-03
Tests

jafors:
Maybe wait for reraised errors - I'll commit here if anything changes in test behaviour.


.gitignore
```diff
@@ -1,129 +1,4 @@
...
-.mypy_cache/
-.dmypy.json
-dmypy.json
-
-# Pyre type checker
-.pyre/
+poetry.lock
+vembrane.egg-info/*
+tests/__pycache__/*
+vembrane/__pycache__/*
```
tedil, line 4: I'd like to keep the .gitignore we already have (and check if it covers the cases you provided here)

jafors: Sorry, somehow there went something wrong. We can remove this here.

tedil: Do you mean the .gitignore or this PR?

jafors: I updated the gitignore with some stuff that wasn't catched last time. Everything should be fine here now

tedil: Thank you! <3

PR #25, opened 2020-07-03
make sure that initialization of known keys with None is not lost

johanneskoester:
Discovered by @mbargull.

### DIFF
--- a/vembrane/__init__.py
+++ b/vembrane/__init__.py
@@ -5,16 +5,9 @@ import math
 import re
 from itertools import chain
 from sys import stderr
-from typing import (
-    Iterator,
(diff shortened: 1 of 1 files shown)


vembrane/__init__.py
```diff
@@ -71,17 +58,17 @@ def filter_vcf(
...
-        env[name] = None
-
     annotation_keys = get_annotation_keys(header)
 
     for record in vcf:
         # setup filter expression env
         env.clear()
+        for name in header.info:
+            env[name] = None
+
```
tedil, line 68: I'm pretty sure we already had that at some point :sweat_smile:

PR #26, opened 2020-07-03
Rename ann

jafors:
Adds flag --rename to rename any desired ANN fields.
Example:
`--rename 'Gene_Name=Gene_Symbol,Annotation_Impact=Impact'`


tedil:
I think this accidentally also includes your changes from the #24 branch

PR #27, opened 2020-07-03
Rename ann

jafors:
As before, a flag to rename ANN fields.
Example:
`--rename 'Gene_Name=Gene_Symbol,Annotation_Impact=Impact'`


vembrane/__init__.py
```diff
@@ -23,10 +23,15 @@
 }
 
 
-def get_annotation_keys(header: VariantHeader,) -> List[str]:
+def get_annotation_keys(header: VariantHeader, rename_dict: dict) -> List[str]:
     for rec in header.records:
         if rec.get("ID") == "ANN":
-            return list(map(str.strip, rec.get("Description").split("'")[1].split("|")))
+            return [
+                rename_dict[ann] if ann in rename_dict else ann
```
tedil, line 30: try `rename_dict.get(ann, ann)` instead!

vembrane/__init__.py
```diff
@@ -130,6 +139,13 @@ def main():
         help="Keep all annotations of a variant if at least one of them passes "
         "the expression.",
     )
+
+    parser.add_argument(
+        "--rename",
+        default="=",
+        help="Replace ANN field names with desired variables. "
+        "Multiple replacements need to be comma-separated.",
+    )
```
tedil, line 148: 1. I'd personally prefer to use the `type` argument of argparse, e.g.:
```python
    def mapping(value: str) -> Dict[str, str]:
        return dict(map(str.strip, v.strip().split("=")) for v in value.split(',') if v)

    parser.add_argument(
        "--rename",
        default="=",
        type=mapping,
        help="Replace ANN field names with desired variables. "
        "Multiple replacements need to be comma-separated.",
    )
```
so that parsing is already done when you access `args.rename`

2. the default being "=" looks weird :D

christopher-schroeder: oh nice, thats really cool

johanneskoester:
Can you explain what purpose this has? Wouldn't that be a separate tool?

vembrane/__init__.py
```diff
@@ -143,6 +152,17 @@ def main():
...
         "the expression.",
     )
+
+    def mapping(value: str) -> Dict[str, str]:
+        return dict(map(str.strip, v.strip().split("=")) for v in value.split(",") if v)
+
+    parser.add_argument(
+        "--rename",
+        default="",
+        type=mapping,
```
johanneskoester, line 162: In any case, the canonical way for mutliple arguments would rather be to define `nargs="+"` here, and thereby expect separation by whitespace.

jafors:
This was just a shot at an idea from @christopher-schroeder , to allow renaming fields for users before writing an expression. I guess this will probably end up implemented in another way.

vembrane/__init__.py
tedil: Either that or multiple `--rename` args. The problem with nargs="+" is that it doesn't work well if you have multiple options/arguments with nargs="+".

tedil:
Closing this for now. I think renaming ANN fields is neat for a single user, but once you start communicating your vembrane command-lines (using new names for the ANN fields) confusion will eventually "arise" ;)

PR #30, opened 2020-07-03
ANN typing

tedil:
For most ANN fields, the type is known in advance, so we can do type conversion. For unknown fields, try auto typing those (the user should be able to opt-out or opt-in, here [not yet implemented]).


tedil:
If we merge this, we should also mention/explain this in the README

PR #34, opened 2020-07-09
Added flag for correct parsing of VEP annotations keys

FelixMoelder:
This PR adds a flag for parsing variant files annotated by VEP.
This is necessary as VEP does not use quotation marks in the description of the ANN-field.
Therefore, the separator get's turned into `:` instead of `'`.
To avoid using flag we could alternatively check if the description contains the string "Ensembl VEP".

The INFO row created bei VEP looks as following:
`##INFO=<ID=ANN,Number=.,Type=String,Description="Consequence annotations from Ensembl VEP. Format: Allele|Consequence|IMPACT|SYMBOL|Gene|Feature_type|Feature|BIOTYPE|EXON|INTRON|HGVSc|HGVSp|cDNA_position|CDS_position|Protein_position|Amino_acids|Codons|Existing_variation|DISTANCE|STRAND|FLAGS|VARIANT_CLASS|SYMBOL_SOURCE|HGNC_ID|CANONICAL|MANE|TSL|APPRIS|CCDS|ENSP|SWISSPROT|TREMBL|UNIPARC|GENE_PHENO|SIFT|PolyPhen|DOMAINS|miRNA|HGVS_OFFSET|AF|AFR_AF|AMR_AF|EAS_AF|EUR_AF|SAS_AF|AA_AF|EA_AF|gnomAD_AF|gnomAD_AFR_AF|gnomAD_AMR_AF|gnomAD_ASJ_AF|gnomAD_EAS_AF|gnomAD_FIN_AF|gnomAD_NFE_AF|gnomAD_OTH_AF|gnomAD_SAS_AF|MAX_AF|MAX_AF_POPS|CLIN_SIG|SOMATIC|PHENO|PUBMED|MOTIF_NAME|MOTIF_POS|HIGH_INF_POS|MOTIF_SCORE_CHANGE|LoFtool">
`

### DIFF
--- a/vembrane/__init__.py
+++ b/vembrane/__init__.py
@@ -74,10 +74,13 @@ class Annotation:
             raise UnknownAnnotation(self._record_idx, ke)
 
 
-def get_annotation_keys(header: VariantHeader,) -> List[str]:
+def get_annotation_keys(header: VariantHeader, vep: bool) -> List[str]:
+    seperator = ":" if vep else "'"
     for rec in header.records:
         if rec.get("ID") == "ANN":
-            return list(map(str.strip, rec.get("Description").split("'")[1].split("|")))
+            return list(
+                map(str.strip, rec.get("Description").split(seperator)[1].split("|"))
+            )
     return []
 
 
@@ -103,12 +106,13 @@ def filter_vcf(
     expression: str,
     ann_key: str = "ANN",
     keep_unmatched: bool = False,
+    vep: bool = False,
 ) -> Iterator[VariantRecord]:
     header = vcf.header
 
     env = dict()
 
-    annotation_keys = get_annotation_keys(header)
+    annotation_keys = get_annotation_keys(header, vep)
 
     for idx, record in enumerate(vcf):
         # setup filter expression env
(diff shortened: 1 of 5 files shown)


vembrane/__init__.py
```diff
@@ -74,10 +74,13 @@ def __getitem__(self, item):
             raise UnknownAnnotation(self._record_idx, ke)
 
 
-def get_annotation_keys(header: VariantHeader,) -> List[str]:
+def get_annotation_keys(header: VariantHeader, vep: bool) -> List[str]:
+    seperator = ":" if vep else "'"
```
johanneskoester, line 78: typo, it should be called `separator`.

tedil:
I'm not a fan of the flag tbh, we should really just check the header for the ANN format (i.e. snpsift vs vep vs whatever) and use the correct annotation parser

FelixMoelder:
I‘ll check on that and replace the flag by checking the header.
Should be more user friendly and prevent Workflows from failing.

vembrane/__init__.py
FelixMoelder: Thanks!

FelixMoelder:
I guess the latest commit is more convenient.
VEP adds information about its version. This appear ahead of the ANN field and can be found just by looking for the "VEP" key.

PR #35, opened 2020-07-09
Optional dot notation

christopher-schroeder:
Just a small change, that allows to also use dot notation to access the fields, so that

`ANN.Annotation_Impact`

instead of
`
ANN["Annotation_Impact"]`

becomes possible. I am aware, that the last should be the default way of doing it, since the name must not be a valid python identifier.

### DIFF
--- a/vembrane/__init__.py
+++ b/vembrane/__init__.py
@@ -37,11 +37,38 @@ globals_whitelist = {
 }
 
 
+class Map(dict):
+    def __init__(self, *args, **kwargs):
+        super(Map, self).__init__(*args, **kwargs)
+        for arg in args:
+            if isinstance(arg, dict):
+                for k, v in arg.iteritems():
+                    self[k] = v
+        if kwargs:
+            for k, v in kwargs.iteritems():
+                self[k] = v
+    def __getattr__(self, attr):
+        return self.get(attr)
+    def __setattr__(self, key, value):
+        self.__setitem__(key, value)
+    def __setitem__(self, key, value):
+        super(Map, self).__setitem__(key, value)
+        self.__dict__.update({key: value})
+    def __delattr__(self, item):
+        self.__delitem__(item)
+    def __delitem__(self, key):
+        super(Map, self).__delitem__(key)
+        del self.__dict__[key]
+
+
 class Sample:
     def __init__(self, record_idx: int, sample: str, format_data: Dict[str, Any]):
         self._record_idx = record_idx
         self._sample = sample
         self._data = format_data
+
+    def __getattr__(self, attr):
+        return self.__getitem__(attr)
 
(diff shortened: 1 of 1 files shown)


vembrane/__init__.py
```diff
@@ -37,12 +37,44 @@
 }
 
 
+class Map(dict):
```
tedil, line 40: why define `Map` if you don't iuse it anywhere? :D

vembrane/__init__.py
```diff
@@ -37,12 +37,44 @@
...
+    def __init__(self, *args, **kwargs):
+        super(Map, self).__init__(*args, **kwargs)
+        for arg in args:
+            if isinstance(arg, dict):
+                for k, v in arg.iteritems():
+                    self[k] = v
+        if kwargs:
+            for k, v in kwargs.iteritems():
+                self[k] = v
```
tedil, line 49: can't you just do e.g. `self.__dict__ = kwargs`?

christopher-schroeder: sorry, the whole Map thing was a first try to just inherit from one class. But then I realized, the easiest way is just adding the 3 functions. I deleted the Map class

christopher-schroeder:
> I'm torn on this issue. On one hand, I like having to type fewer characters, and `.` has only 1 character, while `[""]` has 4. On the other hand, as discussed with Johannes, we'd like to to _not_ add additional confusion (i.e. users exchanging examples with different syntax, or a user expecting to be able to write `ANN.HGVS.c` or `ANN.fancy a"§$nnot` and subsequently failing).
> 
> At the very least, invalid identifiers should be made valid, i.e. by replacing invalid chars with (collapsing) underscore and prepending the identifier with an underscore if the first char happens to be a digit, like so:
> `re.sub(r'\W+|^(?=\d)','_', key)` (see [stackoverflow](https://stackoverflow.com/questions/3303312/how-do-i-convert-a-string-to-a-valid-variable-name-in-python))

At the end, its your rep and you decide. My 5 cent are, that ANN.Annotation_Impact drastically increase the readability and usability and can be applied in the vast majority of the cases. I am skeptic about re.sub(r'\W+|^(?=\d)','_', key), as I think an automatic replacement could lead to massive confusion.

What about an additional replacement.
Example: "Annotation Impact"
valid references are
ANN["Annotation Impact"]
ANN["Annotation_Impact"]
ANN.Annotation_Impact

christopher-schroeder:
We could also add a flag --enable-dot-notation, which then also performs the  re.sub(r'\W+|^(?=\d)','_', key), so that the user is aware, that there is also something and has to opt-in.

tedil:
After some discussion, we've come to the conclusion not to include the dot notation, so as to avoid the problems mentioned in some of the comments in this PR.

PR #37, opened 2020-07-10
Return empty list for list fields

FelixMoelder:
Currently filtering for elements in list fields like `CLIN_SIG` does not perform well in case there is no entry for the field.
This is because we currently return NA if a key has no value/entry.
While this works for most fields expression like `"'unknown_significance' in ANN['CLIN_SIG']"` may fail.

This PR justs includes a quick work around for this specific field but we should implement a more general solution.
Setting up a list containing fields being list-types would be a solution.
But setting up more and more lists and dictionaries which need to be manually managed makes this less robust.

### DIFF
--- a/vembrane/ann_types.py
+++ b/vembrane/ann_types.py
@@ -125,7 +125,6 @@ KNOWN_ANN_TYPE_MAP_SNPEFF = {
         "ERRORS / WARNINGS / INFO",
         lambda x: [v.strip() for v in x.split("/")],
     ),
-    "CLIN_SIG": ("CLIN_SIG", lambda x: [v.strip() for v in x.split("&")]),
 }
 
 KNOWN_ANN_TYPE_MAP_VEP = {
--- a/vembrane/ann_types.py
+++ b/vembrane/ann_types.py
@@ -143,6 +143,7 @@ KNOWN_ANN_TYPE_MAP_VEP = {
     "cDNA_position": ("cDNA_position", str),
     "CDS_position": ("CDS_position", str),
     "Protein_position": ("Protein_position", str),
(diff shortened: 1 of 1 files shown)


vembrane/ann_types.py
```diff
@@ -141,6 +143,7 @@ def __repr__(self):
     "cDNA_position": ("cDNA_position", str),
     "CDS_position": ("CDS_position", str),
     "Protein_position": ("Protein_position", str),
+    "CLIN_SIG": ("CLIN_SIG", lambda x: [v.strip() for v in x.split("&")]),
```
tedil, line 146:
```suggestion
    "CLIN_SIG": ("CLIN_SIG", lambda x: [v.strip() for v in x.split("&")] if x else []),
```

vembrane/ann_types.py
```diff
@@ -68,6 +68,8 @@ def type_ann(
         new_key, value_func = KNOWN_ANN_TYPE_MAP.get(key, (key, try_auto_type))
         return new_key, value_func(value)
     else:
+        if key == "CLIN_SIG":
```
johanneskoester, line 71: I think the correct solution is to add an na_func to the type map. So if value evaluates to false, it can be called instead of value_func.

vembrane/ann_types.py
```diff
@@ -68,6 +68,8 @@ def type_ann(
...
+            return (key, [])
```
tedil, line 72:
```suggestion
```

vembrane/ann_types.py
johanneskoester: Let us turn the type maps into a class, which wraps the dict and has a default na_func that returns na, but can be overwritten by field.

tedil: good idea!

tedil:
Closing since this has been fixed in #38

PR #38, opened 2020-07-13
ANN: Ditch auto typing (default to str), wrap dicts in AnnotationTyper class

tedil:
Instead of trying to auto convert, default to unmodified values, i.e. the user has to do type conversion in the query -- if there is no pre-defined conversion in the `AnnotationTyper`. When no conversion entry is found for a given key, issue an INFO with a link to this project's issue tracker

### DIFF
--- a/vembrane/ann_types.py
+++ b/vembrane/ann_types.py
@@ -102,6 +102,8 @@ class AnnotationEntry:
 
 class DefaultAnnotationEntry(AnnotationEntry):
     def __init__(self, name: str):
+        # TODO issue INFO: No type conversion information about {name},
+        #  please open an issue here: https://github.com/vembrane/vembrane/issues
         super().__init__(name)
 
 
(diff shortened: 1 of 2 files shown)


vembrane/ann_types.py
```diff
@@ -92,55 +53,111 @@ def __init__(self, pos: int, length: int):
...
+    @classmethod
+    def from_vep_str(cls, value: str) -> "PosLength":
+        start, end = (
+            #  the "-" is optional, so vep either has either start and end position
+            [int(v.strip()) for v in value.split("-")]
+            if "-" in value
+            #  or start position only
+            else [int(value.strip())] * 2
+        )
+        return cls(start, start + end)
```
FelixMoelder, line 69: Isn't the second value of the tumple the length and shouldn't it be calculated by `end - start`?

tedil: ... whooops

tedil, line 69:
```suggestion
        return cls(start, end - start)
```

PR #43, opened 2020-07-13
Additional examples

FelixMoelder:


README.md
```diff
@@ -29,11 +29,20 @@ Vembrane allows to simultaneously filter variants based on any `INFO` field, `CH
...
+  vembrane variants.bcf 'INFO["MQ"] == 60'
+  ```  
 * Only keep annotations and variants where consequence contains the word "stream" (matching "upstream" and "downstream"):
   ```
   vembrane variants.vcf 're.search("stream", ANN["Consequence"])'
   ```
 
+* Only keep annotations and variants where CLIN_SIG contains "pathogenic", "likely_pathogenic" or "drug_response":
+  ```
+  vembrane variants.vcf 'any([entry for entry in ["pathogenic", "likely_pathogenic", "drug_response"] if entry in ANN["CLIN_SIG"]])'
```
tedil, line 43: We can use the set literal and its `isdisjoint` method to have a more concise version, like so:
```suggestion
 vembrane variants.vcf 'not {"pathogenic", "likely_pathogenic", "drug_response"}.isdisjoint(ANN["CLIN_SIG"])' 
```
Another way would be something along the lines of (which would look better without the bool and without the set, perhaps we can make CLIN_SIG a set by default?)

```
 vembrane variants.vcf 'bool({"pathogenic", "likely_pathogenic", "drug_response"} & set(ANN["CLIN_SIG"]))' 
```

mbargull: If `ANN["CLIN_SIG"]` is already a `set`:
```python
not ANN["CLIN_SIG"].isdisjoint({"pathogenic", "likely_pathogenic", "drug_response"})
```
If not:
```python
any(entry in ANN["CLIN_SIG"] for entry in ("pathogenic", "likely_pathogenic", "drug_response"))
```
(The latter is probably one of the most natural language-alike, IMHO.)

tedil: That's why I swapped order of sets in my example; if `ANN["CLIN_SIG"]` is not a set, it'll still work, since `set.isdisjoint(foo)` fill turn any iterable `foo` into a set implicitly. But it's true that it isn't as easy to read as the version with `any`.

mbargull: > `set.isdisjoint(foo)` fill turn any iterable `foo` into a set implicitly.

Nope ;).

But we agree on the better readability of the `any` version anyway :tada:.

tedil: > > `set.isdisjoint(foo)` fill turn any iterable `foo` into a set implicitly.
> 
> Nope ;).
> 
> But we agree on the better readability of the `any` version anyway .
Nope nope :D `{'foo', 'bar'}.isdisjoint(['bar'])`

mbargull: Wellp, now I feel really dump. I have no idea how I was misremembering that :facepalm:.
Sorry for ever doubting you, sensei :bow:.

FelixMoelder: So, which optional do we want to use now as example? :D
I just used the `any`-variant to get an example for that (But using `any(entry in ANN["CLIN_SIG"] for entry in ("pathogenic", "likely_pathogenic", "drug_response"))` like @mbargull said as this would be the correct expression).

We could also use both as these are only examples showing how to use the different operations.

tedil: +1 for `any(entry in ANN["CLIN_SIG"] for entry in ("pathogenic", "likely_pathogenic", "drug_response"))`

README.md
```diff
@@ -29,11 +29,20 @@ Vembrane allows to simultaneously filter variants based on any `INFO` field, `CH
   ```
   vembrane variants.bcf 'ANN["Protein_position"] < 10'
   ```
+* Only keep variants where mapping qualitiy is 60:
```
tedil, line 32:
```suggestion
* Only keep variants where mapping quality is exactly 60:
```

README.md
tedil, line 43:
```suggestion
  vembrane variants.vcf 'any(entry in ANN["CLIN_SIG"] for entry in ("pathogenic", "likely_pathogenic", "drug_response"))'
```

PR #46, opened 2020-07-14
Wrap expression str in Expression class, filter ANN only when needed, store ann_key in Expression

tedil:


vembrane/__init__.py
```diff
@@ -108,31 +108,41 @@ def parse_annotation_entry(entry: str,) -> List[str]:
...
-                zip(annotation_keys, parse_annotation_entry(annotation)),
-            )
-        ),
-    )
-    return eval(expression, globals_whitelist, env)
+class Expression:
+    def __init__(self, expression: str, ann_key: str = "ANN"):
+        self._expression = expression
+        self._ann_key = ann_key
+        self._has_ann = f"{ann_key}[" in expression
```
tedil, line 115: atm we just check for the ann_key with "[" appended to determine if the expression actually contains an ANN[ access. However, someone might do something like `"foo" in "".join(ANN)` for whatever reason, in which case it'd fail

vembrane/__init__.py
```diff
@@ -108,31 +108,41 @@ def parse_annotation_entry(entry: str,) -> List[str]:
...
+
+    def annotation_key(self):
+        return self._ann_key
+
+    def filters_annotations(self):
+        return self._has_ann
+
+    def evaluate(
```
tedil, line 123: this is just the eval_expression function moved into Expression and renamed

vembrane/__init__.py
```diff
@@ -162,12 +172,18 @@ def filter_vcf(
...
+            filtered_annotations = [
+                annotation
+                for annotation in annotations
+                if expression.evaluate(idx, annotation, annotation_keys, env,)
+            ]
+        else:
+            annotations = [""]
+            filtered_annotations = (
+                [""] if expression.evaluate(idx, "", annotation_keys, env,) else []
+            )
```
tedil, line 186: if there's no usage of ann_key in the expression, we do not need to re-evaluate the expression for every single annotation in the record but only need to do that exactly once

vembrane/__init__.py
```diff
@@ -162,12 +172,18 @@ def filter_vcf(
...
         env["SAMPLES"] = list(record.samples)
 
-        annotations = dict(record.info).get(ann_key, [""])
-        filtered_annotations = [
-            annotation
-            for annotation in annotations
-            if eval_expression(expression, idx, annotation, annotation_keys, env,)
-        ]
+        if expression.filters_annotations():
+            annotations = dict(record.info).get(ann_key, [""])
```
FelixMoelder, line 176: Do we need this empty string here? Should not be an empty list enough?

vembrane/__init__.py
```diff
@@ -162,12 +172,18 @@ def filter_vcf(
...
+            filtered_annotations = [
+                annotation
+                for annotation in annotations
+                if expression.evaluate(idx, annotation, annotation_keys, env,)
+            ]
+        else:
+            annotations = [""]
```
FelixMoelder, line 183: `annotations` gets initialized but is never used.

vembrane/__init__.py
FelixMoelder, line 176:
```suggestion
            annotations = dict(record.info).get(ann_key, [])
```

vembrane/__init__.py
FelixMoelder, line 183:
```suggestion
```

tedil: it does get used further down!

vembrane/__init__.py
tedil: Yes, if there's an ann_key in the expression but there's no ANN field in the record, the list comprehension still needs to run at least once

tedil: But I guess we can restructure code logic here so it isn't as confusing

FelixMoelder: > Yes, if there's an ann_key in the expression but there's no ANN field in the record, the list comprehension still needs to have at least one entry (i.e. expression.evaluate needs to be called once, which would not happen if the list was empty)

Indeed! :)

PR #49, opened 2020-07-14
Github workflow for building the vembrane dists and publishing them to PyPI

tedil:


.github/workflows/pypi.yml
```diff
@@ -0,0 +1,29 @@
...
+      - uses: actions/checkout@v2
+      - name: Set up Python 3.7
+        uses: actions/setup-python@v1
+        with:
+          python-version: 3.7
+      - name: Install dependencies
+        run:  |
+          python -m pip install --upgrade pip
+          python -m pip install poetry black flake8
+          poetry install --no-interaction
```
mbargull, line 21:
```suggestion
          python -m pip install poetry
```
Should suffice, right?

.github/workflows/pypi.yml
```diff
@@ -0,0 +1,29 @@
...
+jobs:
+  build-and-publish:
+    name: build and publish python dists to PyPI
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Python 3.7
+        uses: actions/setup-python@v1
+        with:
+          python-version: 3.7
```
mbargull, line 16: Since you build `wheel`s: Is it intentional to only build for 3.7 and not 3.8?

mbargull:
BTW, https://github.com/vembrane/vembrane/blob/c859e3d3b1e980bdc08e37cff082f6ec18b1a161/.github/workflows/main.yml#L6 lists `master`, not `main`, i.e., currently only PRs run the other CI job.

.github/workflows/pypi.yml
tedil: tbh I have no clue what I'm doing w.r.t. python packaging

.github/workflows/pypi.yml
tedil: yes, indeed! copied that from the main.yml

tedil:
> BTW,
> 
> https://github.com/vembrane/vembrane/blob/c859e3d3b1e980bdc08e37cff082f6ec18b1a161/.github/workflows/main.yml#L6
> 
> lists `master`, not `main`, i.e., currently only PRs run the other CI job.

I think github has an alias for main <-> master, but we should change that.

.github/workflows/pypi.yml
mbargull: > I have no clue what I'm doing w.r.t. python packaging

I can't remember ever hearing anyone say the opposite about themselves! I'm pretty sure we're all in the same boat even if some tell differently XD

tedil: I think we can restrict poetry to build either sdist or wheel only via the `--format` option, if that helps?

tedil: Or do we have to include multiple python versions in this step?

mbargull: Nvm, sorry, this builds universal wheels and as such doesn't need multiple runs for different versions.

PR #51, opened 2020-07-14
Manage version via pyproject.toml only

tedil:


mbargull:
Alternative: https://github.com/mtkennerly/poetry-dynamic-versioning

PR #61, opened 2020-07-16
Access and parse only those fields that appear in the filter expression

tedil:
Previously, every available bit of information was put into the env before evaluation the expression. Since that involved a lot of string parsing, that was quite some overhead.
So instead we have an `Environment` instance which during its initialization walks the AST of the expression to get all symbol names and exposes an `update(record)` method which only updates those fields that were encountered during walking the AST.
In some testcases for ~100MB bcfs, running time differences were: ~29s (main branch), ~16s (this branch) for a simple expression using one `ANN` and one `INFO` field.

(I think I will have to look at FORMAT in more detail, since I do not have a clue when or why that one is used; i.e. this branch here might discard too many samplenames or so)

### DIFF
--- a/vembrane/__init__.py
+++ b/vembrane/__init__.py
@@ -249,9 +249,6 @@ class Environment:
             "SAMPLES",
         } & names
         available_symbols = globals_whitelist.keys() & names
-        self.globals_whitelist = {
-            key: globals_whitelist[key] for key in available_symbols
-        }
 
         ann_field_name = expression.annotation_key()
         annotation_keys = get_annotation_keys(vcf_header, ann_field_name)
@@ -312,10 +309,10 @@ class Environment:
             ann_field_name: self.annotation,
         }
 
-        # We use self.globals + self.func as a closure.
-        # Do not reassign self.globals, but use .clear()/.update() on it!
-        self.globals = {}
-        self.func = eval(f"lambda: {expression.raw_expression}", self.globals, {})
+        # We use self._globals + self.func as a closure.
+        # Do not reassign self._globals, but use .update() on it!
+        self._globals = {key: globals_whitelist[key] for key in available_symbols}
+        self.func = eval(f"lambda: {expression.raw_expression}", self._globals, {})
 
     def update(self, idx: int, record: VariantRecord):
         self.idx = idx
@@ -328,9 +325,7 @@ class Environment:
         self.annotation.update(self.idx, annotation)
 
     def evaluate(self) -> bool:
-        self.globals.clear()
-        self.globals.update(self.env)
-        self.globals.update(self.globals_whitelist)
+        self._globals.update(self.env)
         return self.func()
 
     def evaluate_with_ann(self, annotation: str) -> bool:
--- a/vembrane/__init__.py
+++ b/vembrane/__init__.py
@@ -387,7 +387,7 @@ def statistics(
     counter = defaultdict(lambda: defaultdict(lambda: 0))
     for record in records:
         for annotation in record.info[ann_key]:
-            for key, value in zip(annotation_keys, parse_annotation_entry(annotation)):
+            for key, value in zip(annotation_keys, split_annotation_entry(annotation)):
                 if value:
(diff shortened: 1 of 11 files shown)


mbargull:
I only skimmed this for now. I'm +1 for lazy evaluation and I'm intrigued by the `names`/`ast.walk` idea. However, the latter can impose usability restrictions, e.g., `(SAMPLES["common-prefix-" + x] for x in ["one", "two"])` might not work then, right?
I believe it's not justifiable to impose usage restrictions for speed improvements in this case. But I think we can get away with something simpler that does not impose restrictions like eval only on dict-lookup. I'll have a look later today.

tedil:
> I only skimmed this for now. I'm +1 for lazy evaluation and I'm intrigued by the `names`/`ast.walk` idea. However, the latter can impose usability restrictions, e.g., `(SAMPLES["common-prefix-" + x] for x in ["one", "two"])` might not work then, right?
> I believe it's not justifiable to impose usage restrictions for speed improvements in this case. But I think we can get away with something simpler that does not impose restrictions like eval only on dict-lookup. I'll have a look later today.

Hm, well, I guess we *can* make such restrictions since some use cases are much more common than others. I mean, vembrane already is way more flexible than other tools, so some slight restrictions wouldn't really hurt, in my opinion.

Since we have the ast, however, we might aswell check for dictionary lookups (there's an AST node type for that!) and look at what's between the brackets. But you can't catch everything.

Or have a "--complex" flag which doesn't do this performance optimization.

mbargull:
I've added a commit (unrelated/orthogonal to your `ast.walk` work -- I could also send that as a separate PR) that shaves off about 60 % runtime in a test case I ran locally. Please do test ;)
Though I still believe we should go with a simpler route on this. I half-ass hacked on the code to see how far a simple "eval only on dict-lookup" will bring us. That showed about 80-90 % of the improvements you achieved, so seems viable. But I haven't had the time to polish it, yet...

vembrane/__init__.py
```diff
@@ -111,77 +180,187 @@ def get_annotation_keys(header: VariantHeader, ann_key: str) -> List[str]:
...
+
+        # Build the env dict used for `eval`
+        self.env = {
+            "INFO": self.info,
+            "FORMAT": self.formats,
+            ann_field_name: self.annotation,
+        }
+
+        self.globals = {}
+        self.func = eval(f"lambda: {expression.raw_expression}", self.globals, {})
```
tedil, line 316: if you run `eval` here, does it actually keep the *reference* of `self.globals`? I mean, since the tests don't fail, it must be that way :D

tedil: In any case, I obviously wasn't aware of the precise `eval` semantics and thought we had to call that everytime something changes, even though the expression (and hence the "compilation") itself does not change.
Very neat trick ;)

Just gave this a try, with your change running time went down to 5s! Awesome!

mbargull: Ah, should've made a comment!
Yep, I use `globals` as a closure here. Hence, we should never reassign `self.globals = ...` and also inject `self.env` into `self.globals` and we're set :).

tedil:
> Though I still believe we should go with a simpler route on this.
i.e. not using the info from the AST? Since the env only gets set up once (i.e. we need to look at the expression only once) and vcf/bcf files may get pretty large, I'm still in favour of checking if the dict lookups are "simple" (i.e. lookup happens via a constant key), and if that's the case, restrict the fields -- otherwise just dump everything we have into `self.globals`.

vembrane/__init__.py
tedil: > and also inject `self.env` into `self.globals` and we're set :).

but you already do that in `evaluate`: `self.globals.update(self.env)`, right?

mbargull: Exactly, I was referring to that line ;)

tedil:
Closing in favour of #63

tedil:
Closing in favour of #63

PR #63, opened 2020-07-20
Parse info/annotation entries only on dict lookup

mbargull:
This is an alternative to gh-61.
Whereas gh-61 does preprocessing via `ast.walk`, this PR does less preprocessing and decides on what values to parse on dicitonary lookups from the expression.
gh-61 should be faster but might prevent some "more exotic" uses/expression. This PR is more flexible but likely slower (and adds less logic).

This is likely functional but needs some cleanup still.

### DIFF
--- a/vembrane/__init__.py
+++ b/vembrane/__init__.py
@@ -77,12 +77,18 @@ class Info:
     def __init__(self, record_idx: int, info_dict: Dict[str, Dict[str, Any]]):
         self._record_idx = record_idx
         self._info_dict = info_dict
+        self._typed_dict = {}
 
     def __getitem__(self, item):
         try:
-            return self._info_dict[item]
-        except KeyError as ke:
-            raise UnknownInfoField(self._record_idx, ke)
+            return self._typed_dict[item]
+        except KeyError:
+            try:
+                untyped_value = self._info_dict[item]
+            except KeyError as ke:
+                raise UnknownInfoField(self._record_idx, ke)
+            value = self._typed_dict[item] = type_info(untyped_value)
+            return value
 
 
 class Annotation:
@@ -172,8 +178,7 @@ def filter_vcf(
         env["QUAL"] = type_info(record.qual)
         env["FILTER"] = record.filter
         env["INFO"] = Info(
-            idx,
-            {key: type_info(record.info[key]) for key in record.info if key != ann_key},
+            idx, {key: record.info[key] for key in record.info if key != ann_key},
         )
 
         formats = {
--- a/vembrane/__init__.py
+++ b/vembrane/__init__.py
@@ -13,7 +13,6 @@ import math
 import re
 import sys
 from collections import defaultdict
-from functools import lru_cache
 from itertools import chain
 from sys import stderr
 from typing import Iterator, List, Tuple
@@ -122,7 +121,7 @@ class Annotation:
     def update(self, record_idx: int, annotation: str):
         self._record_idx = record_idx
         self._data.clear()
-        self._annotation_data = parse_annotation_entry(annotation)
+        self._annotation_data = split_annotation_entry(annotation)
 
     def __getitem__(self, item):
         try:
@@ -150,8 +149,7 @@ def get_annotation_keys(header: VariantHeader, ann_key: str) -> List[str]:
     return []
 
 
-@lru_cache(maxsize=32)
-def parse_annotation_entry(entry: str,) -> List[str]:
+def split_annotation_entry(entry: str,) -> List[str]:
     return entry.split("|")
 
 
@@ -305,7 +303,7 @@ def statistics(
     for record in records:
         for annotation in record.info[ann_key]:
             for key, raw_value in zip(
-                annotation_keys, parse_annotation_entry(annotation)
+                annotation_keys, split_annotation_entry(annotation)
             ):
                 value = raw_value.strip()
                 if value:
--- a/vembrane/__init__.py
+++ b/vembrane/__init__.py
@@ -160,6 +160,7 @@ def filter_vcf(
     annotation_keys = get_annotation_keys(header, ann_key)
 
     for idx, record in enumerate(vcf):
+        yield record
         continue
         # setup filter expression env
         env.clear()
--- a/vembrane/__init__.py
+++ b/vembrane/__init__.py
@@ -272,7 +272,10 @@ def filter_vcf(
             # if the expression contains a reference to the ANN field
             # get all annotations from the record.info field
             # (or supply an empty ANN value if the record has no ANN field)
-            annotations = record.info.get(ann_key, [""])
+            try:
+                annotations = record.info[ann_key]
+            except KeyError:
+                annotations = [""]
             #  … and only keep the annotations where the expression evaluates to true
             filtered_annotations = [
                 annotation for annotation in annotations if env.evaluate(annotation)
--- a/vembrane/__init__.py
+++ b/vembrane/__init__.py
@@ -160,8 +160,6 @@ def filter_vcf(
     annotation_keys = get_annotation_keys(header, ann_key)
 
     for idx, record in enumerate(vcf):
-        yield record
-        continue
         # setup filter expression env
         env.clear()
         for name in header.info:
@@ -192,7 +190,7 @@ def filter_vcf(
             # if the expression contains a reference to the ANN field
             # get all annotations from the record.info field
             # (or supply an empty ANN value if the record has no ANN field)
-            annotations = dict(record.info).get(ann_key, [""])
+            annotations = record.info.get(ann_key, [""])
             #  … and only keep the annotations where the expression evaluates to true
             filtered_annotations = [
                 annotation
--- a/vembrane/__init__.py
+++ b/vembrane/__init__.py
@@ -16,10 +16,11 @@ from collections import defaultdict
 from functools import lru_cache
 from itertools import chain
 from sys import stderr
-from typing import Iterator, List
+from typing import Iterator, List, Tuple
 
 import yaml
 from pysam import VariantFile, VariantRecord, VariantHeader
+from pysam.libcbcf import VariantRecordInfo, VariantRecordSample
 
 from vembrane.ann_types import NA, type_info, ANN_TYPER
 from vembrane.errors import (
@@ -49,7 +50,7 @@ globals_whitelist = {
 
 
 class Sample:
-    def __init__(self, record_idx: int, sample_name: str, sample):
+    def __init__(self, record_idx: int, sample_name: str, sample: VariantRecordSample):
         self._record_idx = record_idx
         self._sample_name = sample_name
         self._sample = sample
@@ -68,7 +69,7 @@ class Sample:
 
 
 class Format:
-    def __init__(self, record_idx: int, record_samples):
+    def __init__(self, record_idx: int, record_samples: List[VariantRecordSample]):
         self._record_idx = record_idx
         self._record_samples = record_samples
         self._sample_formats = {}
@@ -87,7 +88,7 @@ class Format:
 
 
 class Info:
-    def __init__(self, record_idx: int, record_info, ann_key: str):
+    def __init__(self, record_idx: int, record_info: VariantRecordInfo, ann_key: str):
         self._record_idx = record_idx
         self._record_info = record_info
         self._ann_key = ann_key
@@ -210,7 +211,7 @@ class Environment(dict):
         self._globals["ID"] = value
         return value
 
-    def _get_ref_alt(self):
+    def _get_ref_alt(self) -> Tuple[str, List[str]]:
         ref, *alt = chain(self.record.alleles)
         self._globals["REF"], self._globals["ALT"] = ref, alt
         return ref, alt
@@ -218,10 +219,10 @@ class Environment(dict):
     def _get_ref(self) -> str:
         return self._get_ref_alt()[0]
 
-    def _get_alt(self):
+    def _get_alt(self) -> List[str]:
         return self._get_ref_alt()[1]
 
-    def _get_qual(self):
+    def _get_qual(self) -> float:
         value = type_info(self.record.qual)
         self._globals["QUAL"] = value
         return value
@@ -241,7 +242,7 @@ class Environment(dict):
         self._globals["FORMAT"] = value
         return value
 
-    def _get_samples(self):
+    def _get_samples(self) -> List[VariantRecordSample]:
         value = list(self.record.samples)
         self._globals["SAMPLES"] = value
         return value
--- a/vembrane/__init__.py
+++ b/vembrane/__init__.py
@@ -16,7 +16,7 @@ from collections import defaultdict
 from functools import lru_cache
 from itertools import chain
 from sys import stderr
-from typing import Any, Callable, Dict, Iterator, List, Tuple
+from typing import Any, Dict, Iterator, List
 
 import yaml
 from pysam import VariantFile, VariantRecord, VariantHeader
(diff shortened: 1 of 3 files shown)


mbargull:
`False` | `True` | `sum(INFO['PROB_PRESENT']) > 0.0 and ANN['IMPACT'] == 'HIGH'` | commit
-----|------|------|---
 0.2 |  0.7 |  2.9 | branch reduce_lookups 6511279746a3e1164acf998208b46e7380d07853
 1.1 |  1.6 | 27.0 | branch main 48d0bffb786ea54b200ad219d13fb4643d3636ae
 0.2 |  0.2 |  0.2 | baseline read 0e2ec6bc76a7d3b79146859170413b2048ec6d1f
 0.6 |  0.6 |  0.6 | baseline write 76d62127c631c66edd08270369a6e8900c8810b6
 1.0 |  1.5 | 21.0 | closure 2a9da558f8e755d856ff319a93f46dbd313ba5ea
 1.0 |  1.5 | 20.5 | record.info.get 2a0785018f7468c38a7b63d83bd9992f2e029ccd
 0.9 |  1.4 | 20.4 | lazy cast info 83a776b2804b59b3f5f88a969d5efe77b1a4adc0
 0.8 |  1.3 | 20.1 | Environment 7ba3462f51c2179b5e4bbb613d18a0b59082c90c
 0.7 |  1.2 |  5.5 | lazy cast annotation 7d4aa59a581eafcb4d7ac6996709aece00ca903d

tedil:
That's a great overview, thank you!

> `False`	`True`	`sum(INFO['PROB_PRESENT']) > 0.0 and ANN['IMPACT'] == 'HIGH'`	commit
> 0.2	0.7	2.9	branch reduce_lookups [6511279](https://github.com/vembrane/vembrane/commit/6511279746a3e1164acf998208b46e7380d07853)
> 1.1	1.6	27.0	branch main [48d0bff](https://github.com/vembrane/vembrane/commit/48d0bffb786ea54b200ad219d13fb4643d3636ae)
> 0.2	0.2	0.2	baseline read [0e2ec6b](https://github.com/vembrane/vembrane/commit/0e2ec6bc76a7d3b79146859170413b2048ec6d1f)
> 0.6	0.6	0.6	baseline write [76d6212](https://github.com/vembrane/vembrane/commit/76d62127c631c66edd08270369a6e8900c8810b6)
> 1.0	1.5	21.0	closure [2a9da55]

(https://github.com/vembrane/vembrane/commit/2a9da558f8e755d856ff319a93f46dbd313ba5ea)
> 1.0	1.5	20.5	record.info.get [2a07850]

I think we encountered some problems with pysams `VariantInfo.get` which is why we have that seemingly unnecessary `dict` wrapping in the code, but I hope that was just fluke and we can remove that `dict` call.

(https://github.com/vembrane/vembrane/commit/2a0785018f7468c38a7b63d83bd9992f2e029ccd)
> 0.9	1.4	20.4	lazy cast info [83a776b]

As expected, this doesn't shave off much, since info is already parsed/typed, type_info just maps None to NA and wraps tuples in a tuple class.

(https://github.com/vembrane/vembrane/commit/83a776b2804b59b3f5f88a969d5efe77b1a4adc0)
> 0.8	1.3	20.1	Environment [7ba3462]

(https://github.com/vembrane/vembrane/commit/7ba3462f51c2179b5e4bbb613d18a0b59082c90c)
> 0.7	1.2	5.5	lazy cast annotation [7d4aa59](https://github.com/vembrane/vembrane/commit/7d4aa59a581eafcb4d7ac6996709aece00ca903d)

This is great! As expected, parsing the annotation (if there is any, that is) is what takes most of the time. It's still slightly slower than the AST parsing approach, but good enough, I think.

mbargull:
> I think we encountered some problems with pysams `VariantInfo.get`

To avoid any potential issues: When cleaning up things, we can move the `env["INFO"] = ...` and `env[ann_key] = ...` back together and then do a `dict` cast upfront.

mbargull:
Need to adjust tests.

mbargull:
re `VariantRecordInfo.get`: I diff'ed https://github.com/pysam-developers/pysam/blob/v0.16.0.1/pysam/libcbcf.pyx#L2657-L2685 against https://github.com/pysam-developers/pysam/blob/v0.16.0.1/pysam/libcbcf.pyx#L2545-L2575 and they are the sample (apart from raising error/returning default). So, since we use the latter, we should also be fine with using the former.

mbargull:
Runtime is now on par with `reduced_lookups` branch.
This got a little copy-paste heavy, esp. the `_get_<field>` methods...
Feel free to change whatever however you see fit!

mbargull:
> re `VariantRecordInfo.get`: I diff'ed https://github.com/pysam-developers/pysam/blob/v0.16.0.1/pysam/libcbcf.pyx#L2657-L2685 against https://github.com/pysam-developers/pysam/blob/v0.16.0.1/pysam/libcbcf.pyx#L2545-L2575 and they are the sample (apart from raising error/returning default). So, since we use the latter, we should also be fine with using the former.

I was just about to close the diff window I had still open... IDK, I'm probably too tired, but I really didn't noticed that `__getitem__` has https://github.com/pysam-developers/pysam/blob/v0.16.0.1/pysam/libcbcf.pyx#L2562-L2563 which `get` does not have! Seems safer to go the `__getitem__` route then?

vembrane/__init__.py
```diff
@@ -49,52 +49,91 @@
 
 
 class Sample:
-    def __init__(self, record_idx: int, sample: str, format_data: Dict[str, Any]):
+    def __init__(self, record_idx: int, sample_name: str, sample):
```
tedil, line 52: We should type annotate everything, `sample` is something along the lines of `Dict[str, whatever]`

vembrane/__init__.py
```diff
@@ -49,52 +49,91 @@
...
-        except KeyError as ke:
-            raise UnknownSample(self._record_idx, ke)
+        except KeyError:
+            try:
+                record_sample = self._record_samples[item]
+            except KeyError as ke:
+                raise UnknownSample(self._record_idx, ke)
+            sample = Sample(self._record_idx, item, record_sample)
+            self._sample_formats[item] = sample
+            return sample
```
tedil, line 86: Come to think of it, I'd like to switch order of FORMAT access to match other APIs, i.e. `FORMAT[format_field][sample]`

vembrane/__init__.py
```diff
@@ -49,52 +49,91 @@
...
 
 
 class Info:
-    def __init__(self, record_idx: int, info_dict: Dict[str, Dict[str, Any]]):
+    def __init__(self, record_idx: int, record_info, ann_key: str):
```
tedil, line 90:
```suggestion
    def __init__(self, record_idx: int, record_info: Dict[str, str], ann_key: str):
```

vembrane/__init__.py
```diff
@@ -112,86 +151,134 @@ def get_annotation_keys(header: VariantHeader, ann_key: str) -> List[str]:
 
 @lru_cache(maxsize=32)
 def parse_annotation_entry(entry: str,) -> List[str]:
```
tedil, line 153: since all this does is split, it should not be named parse :D

vembrane/__init__.py
```diff
@@ -112,86 +151,134 @@ def get_annotation_keys(header: VariantHeader, ann_key: str) -> List[str]:
...
-            idx,
-            dict(
-                map(
-                    lambda v: ANN_TYPER.convert(v[0], v[1]),
-                    zip(annotation_keys, parse_annotation_entry(annotation)),
-                )
-            ),
-        )
-        return eval(self._expression, globals_whitelist, env)
+    def update(self, idx, record):
```
tedil, line 193:
```suggestion
    def update(self, idx: int, record: VariantRecord):
```

tedil: ...or we monkey patch VariantRecord to have an associated idx

vembrane/__init__.py
```diff
@@ -112,86 +151,134 @@ def get_annotation_keys(header: VariantHeader, ann_key: str) -> List[str]:
...
+        self.idx = idx
+        self.record = record
+        self._globals.update(self._empty_globals)
+
+    def _get_chrom(self):
```
tedil, line 198:
```suggestion
    def _get_chrom(self) -> str:
```

vembrane/__init__.py
```diff
@@ -112,86 +151,134 @@ def get_annotation_keys(header: VariantHeader, ann_key: str) -> List[str]:
...
+        value = self.record.chrom
+        self._globals["CHROM"] = value
+        return value
+
+    def _get_pos(self):
```
tedil, line 203:
```suggestion
    def _get_pos(self) -> int:
```

vembrane/__init__.py
```diff
@@ -112,86 +151,134 @@ def get_annotation_keys(header: VariantHeader, ann_key: str) -> List[str]:
...
+        value = self.record.pos
+        self._globals["POS"] = value
+        return value
+
+    def _get_id(self):
```
tedil, line 208:
```suggestion
    def _get_id(self) -> str:
```

vembrane/__init__.py
```diff
@@ -112,86 +151,134 @@ def get_annotation_keys(header: VariantHeader, ann_key: str) -> List[str]:
...
+        value = self.record.id
+        self._globals["ID"] = value
+        return value
+
+    def _get_ref_alt(self):
```
tedil, line 213:
```suggestion
    def _get_ref_alt(self) -> Tuple[str, List[str]]:
```

vembrane/__init__.py
```diff
@@ -112,86 +151,134 @@ def get_annotation_keys(header: VariantHeader, ann_key: str) -> List[str]:
...
+        ref, *alt = chain(self.record.alleles)
+        self._globals["REF"], self._globals["ALT"] = ref, alt
+        return ref, alt
+
+    def _get_ref(self):
```
tedil, line 218:
```suggestion
    def _get_ref(self) -> str:
```

vembrane/__init__.py
```diff
@@ -112,86 +151,134 @@ def get_annotation_keys(header: VariantHeader, ann_key: str) -> List[str]:
...
+        return self._get_ref_alt()[0]
+
+    def _get_alt(self):
```
tedil, line 221:
```suggestion
    def _get_alt(self) -> List[str]:
```

vembrane/__init__.py
```diff
@@ -112,86 +151,134 @@ def get_annotation_keys(header: VariantHeader, ann_key: str) -> List[str]:
...
+        return self._get_ref_alt()[1]
+
+    def _get_qual(self):
```
tedil, line 224: I guess it's a float?
```suggestion
    def _get_qual(self) -> float:
```

vembrane/__init__.py
```diff
@@ -112,86 +151,134 @@ def get_annotation_keys(header: VariantHeader, ann_key: str) -> List[str]:
...
+        value = type_info(self.record.qual)
+        self._globals["QUAL"] = value
+        return value
+
+    def _get_filter(self):
```
tedil, line 229:
```suggestion
    def _get_filter(self) -> str:
```

vembrane/__init__.py
```diff
@@ -112,86 +151,134 @@ def get_annotation_keys(header: VariantHeader, ann_key: str) -> List[str]:
...
+        value = Info(self.idx, self.record.info, self._ann_key)
+        self._globals["INFO"] = value
+        return value
+
+    def _get_format(self):
+        value = Format(self.idx, self.record.samples)
+        self._globals["FORMAT"] = value
+        return value
+
+    def _get_samples(self):
```
tedil, line 244:
```suggestion
    def _get_samples(self) -> List[str]:
```

vembrane/__init__.py
```diff
@@ -112,86 +151,134 @@ def get_annotation_keys(header: VariantHeader, ann_key: str) -> List[str]:
...
+        value = self.record.filter
+        self._globals["FILTER"] = value
+        return value
+
+    def _get_info(self):
+        value = Info(self.idx, self.record.info, self._ann_key)
+        self._globals["INFO"] = value
+        return value
+
+    def _get_format(self):
```
tedil, line 239:
```suggestion
    def _get_format(self) -> Format:
```

vembrane/__init__.py
```diff
@@ -112,86 +151,134 @@ def get_annotation_keys(header: VariantHeader, ann_key: str) -> List[str]:
...
+        value = type_info(self.record.qual)
+        self._globals["QUAL"] = value
+        return value
+
+    def _get_filter(self):
+        value = self.record.filter
+        self._globals["FILTER"] = value
+        return value
+
+    def _get_info(self):
```
tedil, line 234:
```suggestion
    def _get_info(self) -> Info:
```

vembrane/__init__.py
```diff
@@ -112,86 +151,134 @@ def get_annotation_keys(header: VariantHeader, ann_key: str) -> List[str]:
...
+
+    def __getitem__(self, item):
+        if item == self._ann_key:
+            return self._annotation
+        value = self._globals[item]
+        if value is UNSET:
+            value = self._getters[item]()
+        return value
+
+    def evaluate(self, annotation: str,) -> bool:
```
tedil, line 257: We could make annotation a kwarg with default ""

vembrane/__init__.py
mbargull: Yeah, this is `record.sample`.
I didn't look into this further because when I wanted to add `VariantRecordInfo` as a type I noticed it isn't exported by `pysam`...
Feel free to change this how you see fit!

vembrane/__init__.py
mbargull: Yep, we can do this. In a subsequent PR to not conflate things?

vembrane/__init__.py
mbargull: This is the not-exported `VariantRecordInfo`...

vembrane/__init__.py
mbargull: Be my guest :).
Also, I don't think the LRU cache is really necessary.

vembrane/__init__.py
mbargull: whoops..

vembrane/__init__.py
mbargull: Probably `Tuple[str, Tuple[str]]`

vembrane/__init__.py
tedil: Ah, I see. Well, too bad.

vembrane/__init__.py
mbargull: Looks like `Optional[float]`: https://github.com/pysam-developers/pysam/blob/v0.16.0.1/pysam/libcbcf.pyx#L3117

vembrane/__init__.py
tedil: At least not anymore, since this doesn't do any parsing at all. And who knows if there are any identical annotation entries to start with (we should have checked the VCFs we have before putting this in)

vembrane/__init__.py
mbargull: Ah, we can import those via `pysam.libcbcf` directly.

vembrane/__init__.py
mbargull: Oh, TIL `*x` gives `list` not `tuple` -- you're right :).

vembrane/__init__.py
```diff
@@ -49,52 +49,91 @@
...
+        self._record_idx = -1
+        self._annotation_data = {}
+        self._data = {}
+        annotation_keys = get_annotation_keys(header, ann_key)
+        self._ann_conv = {
+            entry.name: (ann_idx, entry.convert)
+            for ann_idx, entry in enumerate(map(ANN_TYPER.get_entry, annotation_keys))
+        }
+
+    def update(self, record_idx: int, annotation: str):
```
tedil, line 121: This masks `Dict.update`. Is that intentional or should we rename that to something like `update_from` or `update_from_record` or `gather_info` or ...?

vembrane/__init__.py
mbargull: FYI, it looks like the other fields are `Optional` also. Which also include `alleles`, meaning `ref, *alt =` would fail in that case -- but then things seem to be strange in any case..

vembrane/__init__.py
mbargull: Hm, yes we can rename it. Albeit, `dict.update` should never be called :). TBH, it is a little hacky to have `Environment` derive from `dict`. But it has to be `dict` to be able to act as `globals`.
If we find this too conflated going forward, we can refactor things a bit later, maybe.

tedil: Well keep it as is for now.

mbargull: Agreed.

mbargull:
> not having to maintain AST parsing

Yep, but I can imagine it has been a fun experience tinkering with the AST ^^.

PR #65, opened 2020-07-21
Swap order of access for FORMAT from `sample → field` to `field → sample`

tedil:


vembrane/__init__.py
```diff
@@ -48,42 +52,55 @@
...
         except KeyError:
             try:
-                sample_format = self._sample[item]
+                record_sample = self._record_samples[sample]
             except KeyError as ke:
-                raise UnknownFormatField(self._record_idx, self._sample_name, ke)
-            self._data[item] = sample_format
-            return sample_format
+                raise UnknownSample(
+                    self._record_idx, str(ke).split(":")[1][1:].strip("'")
```
mbargull, line 72: What does the `str(ke)` contain and what are you extracting? Something more than just `sample`?

tedil: Yes, this is ugly, but the KeyError's sole arg is `'invalid sample name: SomeUnknownSampleName'`. Oh wow I could have just used `sample` here, nvm, will fix.

vembrane/__init__.py
```diff
@@ -48,42 +52,51 @@
...
             try:
-                sample_format = self._sample[item]
-            except KeyError as ke:
-                raise UnknownFormatField(self._record_idx, self._sample_name, ke)
-            self._data[item] = sample_format
-            return sample_format
+                record_sample = self._record_samples[sample]
+            except KeyError:
+                raise UnknownSample(self._record_idx, sample)
+            value = record_sample[self._name]
```
mbargull, line 72: `record.sample[fmt]` is guaranteed to exist, right?
In the old code we had `{fmt: record.samples[sample][fmt] for fmt in record.format}`, hence I assume it is guaranteed, but just checking/asking :).

tedil: I'm not too sure to be honest. I always assumed all samples need to have every FORMAT field specified in the header (the VCF specification isn't too clear on this).
So at this point in the code, if `record.format[fmt]` didn't already KeyError, `record.samples[sample][fmt]` shouldn't either.

> LGTM (barring a missing test case :P).
We should definitely add more testcases.

PR #66, opened 2020-07-21
Try evaluating the filter expression for the first record before writing the VCF header

tedil:


vembrane/__init__.py
```diff
@@ -401,13 +401,22 @@ def main():
...
 
+        records = filter_vcf(
+            vcf,
+            args.expression,
+            args.annotation_key,
+            keep_unmatched=args.keep_unmatched,
+        )
+
+        try:
+            first_record = next(records)
```
mbargull, line 412:
```suggestion
            first_record = list(islice(records, 1))
```
(and `from itertools import islice`)

mbargull: Otherwise the `chain` below will not work and we `exit(1)` for empty VCFs, which are valid input.

PR #68, opened 2020-07-22
Update README.md

christopher-schroeder:
str, float, int

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,9 +6,11 @@ Vembrane allows to simultaneously filter variants based on any `INFO` field, `CH
 ## Filter expression
 The filter expression can be any valid python expression that evaluates to `bool`. However, functions and symbols available have been restricted to the following:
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -8,7 +8,7 @@ The filter expression can be any valid python expression that evaluates to `bool
 
  * `any`, `all`, `min`, `max`, `sum`
  * `list`, `dict`, `set`, `tuple`,
- * `zip`, `map`
+ * `zip`, `map`, `str`, `float`, `int`
```
tedil, line 11: Perhaps we should group them by "kind" or return type? E.g.:
```suggestion
 * `any`, `all`, 
 * `min`, `max`, `sum`
 * `list`, `dict`, `set`, `tuple`,
 * `zip`, `map`
 * `str`, `float`, `int`
```

PR #70, opened 2020-07-22
remove header.info from globals, add for Info[..]=NA default

mbargull:


vembrane/__init__.py
```diff
@@ -210,8 +221,8 @@ def __init__(self, expression: str, ann_key: str, header: VariantHeader):
             "FORMAT": self._get_format,
             "SAMPLES": self._get_samples,
         }
-        self._empty_globals = {name: NA for name in header.info}
```
mbargull, line 213: IIRC, we said we didn't want the `INFO` fields in `globals`, but didn't removed them ^^.

tedil: yup... :D

PR #71, opened 2020-07-22
add __contains__ method

mbargull:


vembrane/__init__.py
```diff
@@ -68,7 +68,16 @@
...
+class NoValueDict:
+    def __contains__(self, item):
+        try:
+            value = self[item]
+        except KeyError:
+            return False
+        return value is not NA
+
+
+class Format(NoValueDict):
```
mbargull, line 80: Does it makes sense to add this for `Format`/`Formats`? I didn't really think about it.

mbargull: I mean we should add a `__contains__` to those in any case.

tedil: Talking about `Format` opens another can of worms. But I think it doesn't do any harm to have those two classes be NoValueDicts, for the time being.

PR #73, opened 2020-07-24
Only allow one ALT allele per record, error if more than one is encountered

tedil:
In order to make syntax for handling `INFO` fields with one value per allele less unwieldy and reduce confusion, this PR forces users to have input files which have only one ALT allele per record, which can for example be achieved by calling `bcftools norm -m-any foo.vcf`. We could call `bcftools norm` ourselves via pysam (if it weren't for https://github.com/pysam-developers/pysam/issues/506), but that doesn't handle snpeff/vep/… annotations in the ANN/CSQ fields. We could also split multi-allelic records ourselves, but we'd rather not introduce additional functionality/complexity when there are established tools made for exactly this purpose.

~~However, this does not handle `FORMAT` yet, so I'll leave this as a draft for now.~~

### DIFF
--- a/vembrane/__init__.py
+++ b/vembrane/__init__.py
@@ -13,7 +13,7 @@ import sys
 from collections import defaultdict
 from itertools import chain, islice
 from sys import stderr
-from typing import Iterator, List, Set, Tuple
+from typing import Iterator, List, Tuple, Dict
 
 import yaml
 from pysam import VariantFile, VariantRecord, VariantHeader
@@ -23,6 +23,7 @@ from pysam.libcbcf import (
     VariantRecordFormat,
     VariantRecordSamples,
 )
+
 from vembrane.ann_types import NA, type_info, ANN_TYPER
 from vembrane.errors import (
     UnknownAnnotation,
@@ -31,6 +32,7 @@ from vembrane.errors import (
     UnknownFormatField,
     UnknownSample,
     VembraneError,
+    MoreThanOneAltAllele,
 )
 from vembrane.globals_whitelist import globals_whitelist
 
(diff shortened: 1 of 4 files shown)


vembrane/ann_types.py
```diff
@@ -43,11 +45,16 @@ def __getitem__(self, spec):
...
+def type_info(value, number="."):
     if value is None:
         return NA
     if isinstance(value, tuple):
-        return InfoTuple(value)
+        if number != "A":
+            return InfoTuple(value)
+        else:
+            if len(value) != 1:
+                raise MoreThanOneAltAllele()
```
mbargull, line 56: Should this also be raised if `number == "G" and len(value) != 2`?

vembrane/ann_types.py
```diff
@@ -43,11 +45,16 @@ def __getitem__(self, spec):
...
+            return value[0]
```
mbargull, line 57: This branch doesn't do the `None` to `NA` substitution.

tedil: Good catch!

vembrane/ann_types.py
tedil: Only for diploid cases, I guess?

mbargull: Why? Do the programs to split the VCFs only split "A" but not "G"?

tedil: Hm, hopefully they handle that correctly.
We also have to check `number == "R" and len(value) != 2`, right?

mbargull: Oh, whoops, I meant to write "R" not "G"!

tedil: Could be less ugly, but the case `number == "R"` is handled now

vembrane/__init__.py
```diff
@@ -197,7 +209,17 @@ def __init__(self, expression: str, ann_key: str, header: VariantHeader):
             "FORMAT": self._get_format,
             "SAMPLES": self._get_samples,
         }
-        self._header_info_fields = set(header.info)
+
+        self._numbers = {
```
mbargull, line 213: Oh, forgot: This could be just `numbers` if we only use it in this function.

PR #74, opened 2020-07-24
Export NA to enable checks such as `INFO["DP"] is NA`

tedil:


vembrane/globals_whitelist.py
```diff
@@ -6,6 +6,8 @@
 #        *sorted(o for o in dir(__builtins__) if o.islower() and not o.startswith("_")),
 #        sep="\n",
 #    )'
+from vembrane import NA
```
mbargull, line 9:
```suggestion
from .ann_types import NA
```
instead of using the reexport.
General remark: To avoid potential circular dependencies, I'd advice to have `__init__.py` mostly empty apart from exporting version information for a toplevel package and -- if necessary -- just reexport imports from submodules.
For `vembrane`, I'd just put `__version__` into `__init__.py` and the rest in a separate module, e.g., `cli.py`/`main.py`/etc. (whilst adjusting the entrypoint to point to `vembrane.cli:main`).

tedil: I agree

PR #75, opened 2020-07-27
Refactor module structure

tedil:
As prompted by @mbargull in #74, tidying up modules a bit to avoid cyclic imports etc.


vembrane/globals.py
```diff
@@ -102,7 +102,7 @@
     "__package__": None,
 }
 
-globals_whitelist = {
+globals_allowlist = {
```
mbargull, line 105: Since this isn't using an "established compound word" anymore, this would be a good opportunity to use more generic (arguably more readable) snake_case, i.e., `globals_allow_list`.
Or even better: `allowed_global_symbols` or just `allowed_globals` since we don't need the `_list` part.

vembrane/cli.py
```diff
@@ -0,0 +1,184 @@
+import argparse
+import ast
+import sys
+from collections import defaultdict
+from itertools import islice, chain
+from sys import stderr
+from typing import Iterator
+
+import yaml
```
mbargull, line 9: By the by: Can you move `pyyaml = "^5.3"` from the test to the normal dependencies in `pyproject.toml`? I forgot that again and have been reminded just now ^^.

vembrane/globals.py
tedil: yup!

vembrane/cli.py
tedil: sure can do!

mbargull: It is a test dep due to a different reason, right. So, copy not move, like you did and not how I said, is correct :).

PR #80, opened 2020-09-24
Support BNDs

christopher-schroeder:
Support for BND events in filtering


vembrane/modules/filter.py
```diff
@@ -62,46 +62,89 @@ def add_subcommmand(subparsers):
...
         "the expression.",
     )
+    parser.add_argument(
+        "--preserve-order",
+        default=False,
+        action="store_true",
+        help="Make sure that the variant order of the output matches that of the \
+              input. For performance reasons only make use of this option, if the \
+              input might contain breakends (BNDs) as the order of all other variants \
+              are preserved anyway.",
```
tedil, line 72:
```suggestion
        help="Ensures that the order of the output matches that of the \
              input. This is only useful if the input contains breakends (BNDs) \
              since the order of all other variants is preserved anyway.
```

vembrane/modules/filter.py
```diff
@@ -62,46 +62,89 @@ def add_subcommmand(subparsers):
...
+    )
+
+
+def test_and_update_record(env, idx, record, ann_key, keep_unmatched):
```
tedil, line 76: I think we should have type info here as well, especially since the return type is now a tuple of things?

vembrane/modules/filter.py
```diff
@@ -62,46 +62,89 @@ def add_subcommmand(subparsers):
...
+        #  … and only keep the annotations where the expression evaluates to true
+        filtered_annotations = [
+            annotation for annotation in annotations if env.evaluate(annotation)
+        ]  # TODO: if keep_unmatched, that could be "any(iterator)" instead of list
+
+        if not keep_unmatched and (len(annotations) != len(filtered_annotations)):
+            # update annotations if they have actually been filtered
+            record.info[ann_key] = filtered_annotations
+
+        return record, len(filtered_annotations) > 0
```
tedil, line 95: is `len(filtered_annotation) > 0` the correct "pass" bool when `keep_unmatched is True`?

vembrane/modules/filter.py
```diff
@@ -62,46 +62,89 @@ def add_subcommmand(subparsers):
...
+        record, passed = test_and_update_record(
+            env, idx, record, ann_key, keep_unmatched
+        )
+        if passed:
+            is_bnd = "SVTYPE" in info_keys and record.info.get("SVTYPE", None) == "BND"
+            if is_bnd:
+                event = record.info.get("EVENT", None)
+                events.add(event)
+            elif not preserve_order:
+                # if preserver order, we will output everything in the second pass *
```
tedil, line 126:
```suggestion
                # if preserve_order is True, we will output everything in the second pass instead
```

vembrane/modules/filter.py
```diff
@@ -62,46 +62,89 @@ def add_subcommmand(subparsers):
...
-                record.info[ann_key] = filtered_annotations
-            yield record
-        else:
-            # otherwise, the annotations are irrelevant w.r.t. the expression,
-            # so we can omit them
-            if env.evaluate():
+        record, passed = test_and_update_record(
+            env, idx, record, ann_key, keep_unmatched
+        )
+        if passed:
```
tedil, line 120: not sure if "passed" is the most descriptive name here, but I'm not really sure what else to use, either, so... :D

vembrane/modules/filter.py
```diff
@@ -62,46 +62,89 @@ def add_subcommmand(subparsers):
...
+            is_bnd = "SVTYPE" in info_keys and record.info.get("SVTYPE", None) == "BND"
+            if is_bnd:
+                event = record.info.get("EVENT", None)
+                events.add(event)
+            elif not preserve_order:
+                # if preserver order, we will output everything in the second pass *
                 yield record
+
+    if len(events) > 0:
+        # perform a second pass
```
tedil, line 130:
```suggestion
        # perform a second pass if the first pass filtered breakend (BND) events
        # since these are compound events which have to be filtered jointly
```

vembrane/modules/filter.py
```diff
@@ -62,46 +62,89 @@ def add_subcommmand(subparsers):
...
+            event = record.info.get("EVENT", None)
+
+            if is_bnd:
+                if event not in events:
+                    # only bnds with valid event
+                    continue
             else:
-                continue
+                if not preserve_order:
+                    # if preserver order, we will output everything in the second pass *
```
tedil, line 142: isn't this the second pass?

vembrane/modules/filter.py
```diff
@@ -62,46 +62,89 @@ def add_subcommmand(subparsers):
...
+    if len(events) > 0:
+        # perform a second pass
+        vcf.reset()
+        for idx, record in enumerate(vcf):
+            is_bnd = "SVTYPE" in info_keys and record.info.get("SVTYPE", None) == "BND"
+            event = record.info.get("EVENT", None)
+
+            if is_bnd:
+                if event not in events:
+                    # only bnds with valid event
```
tedil, line 138:
```suggestion
                    # only bnds with a valid associated event need to be considered, so skip the rest
```

tedil (changes requested):

vembrane/modules/filter.py
christopher-schroeder: added some type infos

vembrane/modules/filter.py
christopher-schroeder: You are right, I don't like the name either. But I was strugeling finding a suitable one. I renamed it to record_has_passed. It is only used once, so it does not really matter.

vembrane/modules/filter.py
christopher-schroeder: Yes. If all annotations are filtered then the record becomes invalid, regardless of keep_unmatched.
If keep_unmatched is True, the only consequence is, that the record annotations are not updated.

vembrane/modules/filter.py
christopher-schroeder: Yes, I thought It might be a good idea to use the same comment twice and use the * at the end to indicate that the two positions belong together. But I seems that this didn't work :D

tedil (approved):

tedil:
I just realised we didn't update the README yet, should probably do that before merging!

PR #84, opened 2020-10-07
More custom VEP annotation types

tedil:
For some of the annotations I had to guess their possible values and separators, since neither
https://www.ensembl.org/info/docs/tools/vep/online/input.html#ident
nor
https://www.ensembl.org/info/docs/tools/vep/vep_formats.html
nor
https://www.ensembl.org/Help/Glossary?id=346
has any information about them. But I think we're good.

### DIFF
--- a/vembrane/ann_types.py
+++ b/vembrane/ann_types.py
@@ -545,10 +545,12 @@ KNOWN_ANN_TYPE_MAP_VEP = {
         "NB: not available for GRCh37",
     ),
     "TSL": AnnotationEntry(
-        "TSL", description="Transcript support level. " "NB: not available for GRCh37"
+        "TSL", description="Transcript support level. NB: not available for GRCh37"
     ),
-    "PUBMED": AnnotationEntry(
-        "PUBMED", description="Pubmed ID(s) of publications that cite existing variant"
+    "PUBMED": AnnotationListEntry(
+        "PUBMED",
+        description="Pubmed ID(s) of publications that cite existing variant",
+        sep="&",
     ),
     "SOMATIC": AnnotationListEntry(
         "SOMATIC",
@@ -654,7 +656,7 @@ KNOWN_ANN_TYPE_MAP_VEP = {
         typefunc=float,
         description="Provides a rank of genic intolerance and "
         "consequent susceptibility to disease based on the ratio of Loss-of-function "
-        "(LoF)to synonymous mutations.",
+        "(LoF) to synonymous mutations.",
     ),
 }
 
--- a/vembrane/ann_types.py
+++ b/vembrane/ann_types.py
@@ -1,6 +1,7 @@
 from collections import defaultdict
 from sys import stderr
 from typing import Union, Iterable, Tuple, Dict, Callable, Any
+import re
 
 from .errors import MoreThanOneAltAllele
 
@@ -177,10 +178,14 @@ class AnnotationNumberTotalEntry(AnnotationListEntry):
         super().__init__(name, sep="/", inner_typefunc=int, **kwargs)
 
 
+PREDICTION_SCORE_REGEXP: re.Pattern = re.compile(r"(.*)\((.*)\)")
+
+
 class AnnotationPredictionScoreEntry(AnnotationEntry):
     def __init__(self, name: str, **kwargs):
         def typefunc(x: str) -> Dict[str, float]:
-            return {x[: x.rindex("(")]: float(x[x.index("(") + 1 : x.rindex(")")])}
+            match = PREDICTION_SCORE_REGEXP.findall(x)[0]
+            return {match[0]: float(match[1])}
 
         super().__init__(name, typefunc=typefunc, nafunc=lambda: dict(), **kwargs)
 
--- a/vembrane/ann_types.py
+++ b/vembrane/ann_types.py
@@ -25,10 +25,13 @@ class NoValue:
     def __bool__(self):
         return False
 
-    def __repr__(self):
+    def __str__(self):
         # nonexistent fields will result in an empty string
         # should be configurable in upcoming versions
         return ""
+
+    def __repr__(self):
+        return f"{self.__class__.__name__}()"
 
 
 NA = NoValue()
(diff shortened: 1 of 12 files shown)


vembrane/ann_types.py
```diff
@@ -140,12 +141,50 @@ def description(self):
...
+    def __init__(self, name: str, **kwargs):
+        def typefunc(x):
+            key_value_tuples = map(lambda v: v.split(":", maxsplit=1), x.split("&"))
+            mapping = defaultdict(list)
+            for key, value in key_value_tuples:
+                mapping[key].append(value)
+            # make sure non-existent keys do not return list() on __getitem__ calls
+            # in the filter expression (without actually building a new dict from
+            # the defaultdict)
+            mapping.default_factory = None
```
mbargull, line 169: So, it is intentional that a user lookup returns `None` for unknown values? (Just asking -- I didn't look at the surrounding code at all. I.e., I also don't know whether we want `None` or `NA` here.)

vembrane/ann_types.py
```diff
@@ -140,12 +141,50 @@ def description(self):
...
+
+class AnnotationNumberTotalEntry(AnnotationListEntry):
+    def __init__(self, name: str, **kwargs):
+        super().__init__(name, sep="/", inner_typefunc=int, **kwargs)
+
+
+class AnnotationPredictionScoreEntry(AnnotationEntry):
+    def __init__(self, name: str, **kwargs):
+        def typefunc(x: str) -> Dict[str, float]:
+            return {x[: x.rindex("(")]: float(x[x.index("(") + 1 : x.rindex(")")])}
```
mbargull, line 183:
```suggestion
            return {x[: x.index("(")]: float(x[x.index("(") + 1 : x.rindex(")")])}
```
(Or use a regex. Regex is probably a little more readable, but I'm okay with both.)

vembrane/ann_types.py
```diff
@@ -289,14 +324,12 @@ def convert(self, key: str, value: str) -> Tuple[str, AnnotationType]:
...
-    #  "a flag indicating if the variant falls in a high information position of a
-    #  transcription factor binding profile (TFBP)"
-    # "HIGH_INF_POS": AnnotationEntry(
-    #     "HIGH_INF_POS",
-    #     description="A flag indicating if the variant falls in a high information "
-    #                 "position of a transcription factor binding profile (TFBP)",
-    # ),
+    "HIGH_INF_POS": AnnotationEntry(
+        "HIGH_INF_POS",
+        typefunc=lambda x: True if x == "Y" else False,
```
mbargull, line 329:
```suggestion
        typefunc=lambda x: if x == "Y",
```
Though, if you think `True if ... else False` is more readable, I'm happy to leave it as is!
If micro-optimization performance is of interest,
```suggestion
        typefunc="Y".__eq__,
```
;).

vembrane/ann_types.py
```diff
@@ -308,29 +341,25 @@ def convert(self, key: str, value: str) -> Tuple[str, AnnotationType]:
...
-    #  "a flag indicating if the transcript is denoted as the canonical transcript
-    #  for this gene"
-    # "CANONICAL": AnnotationEntry(
-    #     "CANONICAL",
-    #     description="A flag indicating if the transcript is denoted as the canonical "
-    #                 "transcript for this gene",
-    # ),
+    "CANONICAL": AnnotationEntry(
+        "CANONICAL",
+        typefunc=lambda x: True if x == "YES" else False,
```
mbargull, line 346: (Same as above)
```suggestion
        typefunc=lambda x: if x == "YES",
```
Though, if you think `True if ... else False` is more readable, I'm happy to leave it as is!
If micro-optimization performance is of interest,
```suggestion
        typefunc="YES".__eq__,
```
;).

vembrane/common.py
```diff
@@ -32,7 +32,10 @@ def get_annotation_keys(header: VariantHeader, ann_key: str) -> List[str]:
             continue
         if rec.get("ID") == ann_key:
             return list(
-                map(str.strip, rec.get("Description").split(separator)[1].split("|"))
+                map(
+                    str.strip,
+                    rec.get("Description").strip('"').split(separator)[1].split("|"),
```
mbargull, line 37: Is the content of `Description` somewhat well-formed/restricted? Or could it also be something like
`"Description": '"Foo: Yadda yadda "quote""'` <- having an additional `"` before the end double quote.

vembrane/ann_types.py
```diff
@@ -140,12 +141,50 @@ def description(self):
...
+
+        super().__init__(name=name, typefunc=typefunc, **kwargs)
+
+
+class AnnotationNumberTotalEntry(AnnotationListEntry):
+    def __init__(self, name: str, **kwargs):
+        super().__init__(name, sep="/", inner_typefunc=int, **kwargs)
+
+
+class AnnotationPredictionScoreEntry(AnnotationEntry):
```
mbargull, line 180: Is it intentional that these are lists of `int`/`float`? (Again, didn't look at the context here, sorry.)

vembrane/ann_types.py
tedil: Oh well, good point. Probably something akin to the `nafunc` but with one (unneeded) param? Didn't think that one through

vembrane/ann_types.py
tedil: ... and we could compile it in advance, might do that

vembrane/common.py
tedil: it's well formed, as far as I know.

vembrane/ann_types.py
tedil: Yes, basically a pair `[number, total]`. Perhaps a custom type with number and total slots might be better?

vembrane/ann_types.py
tedil: `(.*)\((.*)\)` should do the trick :D

tedil: but is it more readable... I don't know.

christopher-schroeder (approved):

PR #85, opened 2020-10-20
Custom functions

tedil:
`SnpSift` has some "custom" functions for counting the number of homo-/heterozygous samples, as well as checking if a sample's genotype is hom/het/var/ref. So I implemented them for vembrane, too. Not sure if I this implementation is correct and also not quite sure if this is the best way to include such functions.
I could also think of having a better Format/Sample class which allows for nicer and more expressive syntax, e.g. `SAMPLES[0].is_hom()`.

### DIFF
--- a/vembrane/globals.py
+++ b/vembrane/globals.py
@@ -6,6 +6,8 @@ import re
 #        *sorted(o for o in dir(__builtins__) if o.islower() and not o.startswith("_")),
 #        sep="\n",
 #    )'
+from typing import Dict, Any
+
 from .ann_types import NA
 
 _builtins = {
@@ -109,3 +111,60 @@ allowed_globals = {
     **_explicit_clear,
     "NA": NA,
 }
+
+
+def custom_functions(env) -> Dict[str, Any]:
+    return {
+        "count_hom": eval(
+            "lambda: "
+            "sum(all(x == FORMAT['GT'][s][0] for x in FORMAT['GT'][s][1:]) "
+            "for s in SAMPLES)",
+            env,
+            {},
+        ),
+        "count_het": eval(
+            "lambda: "
+            "sum(any(x != FORMAT['GT'][s][0] for x in FORMAT['GT'][s][1:]) "
+            "for s in SAMPLES)",
+            env,
+            {},
+        ),
+        "count_ref": eval(
+            "lambda: "
+            "sum(all(x == 0 for x in FORMAT['GT'][s][1:]) "
+            "for s in SAMPLES)",
+            env,
+            {},
+        ),
+        "count_var": eval(
+            "lambda: "
+            "sum(any(x != 0 for x in FORMAT['GT'][s][1:]) "
+            "for s in SAMPLES)",
+            env,
+            {},
+        ),
+        "is_hom": eval(
+            f"lambda sample: "
+            f"all(x == FORMAT['GT'][sample][0] "
+            f"for x in FORMAT['GT'][sample][1:])",
+            env,
+            {},
+        ),
+        "is_het": eval(
+            f"lambda sample: "
+            f"any(x != FORMAT['GT'][sample][0] "
+            f"for x in FORMAT['GT'][sample][1:])",
+            env,
+            {},
+        ),
+        "is_ref": eval(
+            f"lambda sample: " f"all(x == 0 " f"for x in FORMAT['GT'][sample][1:])",
+            env,
+            {},
+        ),
+        "is_var": eval(
+            f"lambda sample: " f"any(x != 0 " f"for x in FORMAT['GT'][sample][1:])",
+            env,
+            {},
+        ),
+    }
--- a/vembrane/representations.py
+++ b/vembrane/representations.py
@@ -2,8 +2,6 @@ import ast
 from itertools import chain
 from typing import Dict, List, Tuple
 
-from .common import get_annotation_keys, split_annotation_entry
-
 from pysam.libcbcf import (
     VariantRecordSamples,
     VariantRecordFormat,
@@ -13,21 +11,20 @@ from pysam.libcbcf import (
     VariantRecordSample,
 )
 
-from .globals import allowed_globals
-
 from .ann_types import (
     NA,
     type_info,
     ANN_TYPER,
     MoreThanOneAltAllele,
 )
-
+from .common import get_annotation_keys, split_annotation_entry
 from .errors import (
     UnknownSample,
     UnknownFormatField,
     UnknownInfoField,
     UnknownAnnotation,
 )
+from .globals import allowed_globals, custom_functions
 
 
 class NoValueDict:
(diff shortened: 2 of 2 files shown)


christopher-schroeder:
The functions seems to be fine, just a quick question. The functions will probably fail on haploid data, such as sperm, oocytes and bacteria. But I think thats ok.
What about triploid organism and more? Are the terms "heterozygous" and "homozygous" defined for these? Because the functions will work, but maybe not following the definitions.

tedil:
> The functions seems to be fine, just a quick question. The functions will probably fail on haploid data, such as sperm, oocytes and bacteria. But I think thats ok.

`count_hom()` will trivially always be `True` for lists of length 1 (and 0), i.e. haploid data (and… incomplete/erroneous data?) (analogous, `count_het()` will always be `False` in these cases)

> What about triploid organism and more? Are the terms "heterozygous" and "homozygous" defined for these? Because the functions will work, but maybe not following the definitions.

Well, that was my question on discord, too; I didn't find any proper definitions for those two terms, so I stuck with the "homozygous means everything is the same, while heterozygous means there's at least one different". And the way the functions are defined works for these definitions.

christopher-schroeder:
Oh sorry I did not follow the discord discussion today. Seems legit then.

christopher-schroeder (approved):

dlaehnemann:
I think the definition of homozygous as "all alleles identical" and heterozygous as "at least one allele differs", as Sven suggested on Discord, makes sense.

However, does this correctly handle NAs or rather `.` from the specs, or does it si ply count `.` as not matching any allele and thus considers any sample sites with dot allele(s) as heterozygous?

tedil:
> However, does this correctly handle NAs or rather `.` from the specs, or does it si ply count `.` as not matching any allele and thus considers any sample sites with dot allele(s) as heterozygous?

Good catch; I will have to look into this. But then, what does "correctly handle NAs" mean in this context?

dlaehnemann:
Yeah, I'm not sure either. Otherwise I would have suggested something. But I guess fully NA genotypes (i.e. `.` or `.|./.` or similar) shoul result in `false` for both hom and het. But for partially missing (e.g. `.|0`), I don't know...

tedil:
> Yeah, I'm not sure either. Otherwise I would have suggested something. But I guess fully NA genotypes (i.e. `.` or `.|./.` or similar) shoul result in `false` for both hom and het. But for partially missing (e.g. `.|0`), I don't know...

Let's look at an example:
Assuming we have 6 samples with genotypes `./.`, `1/1`, `0/0`, `0/1`, `./0` and `./1` we'd expect/want the following to be true:

- `count_hom() == 2` (because of `1/1` and `0/0`) :heavy_check_mark:
- `count_het() == 1` (because of `0/1`) :x: 
- `count_ref() == 1` (because of `0/0`) :heavy_check_mark:
- `count_var() == 2` (because of `1/1` and `0/1`) :x:

`count_hom` and `count_var` are correct since:
-  `.` is interpreted as `NA` (`NoValue`)
- comparisons with `NA` are always `False`
- they are formulated in terms of `all(gt == first_gt)` and `all(gt == 0)` respectively

Because `count_het` and `count_ref` are formulated in terms of `any(gt != first_gt)` and `any(gt != 0)` respectively, the `NA` comparison lead to "false positives". An easy fix would hence be to skip the NAs
`count_het := sum(any(x != FORMAT['GT'][s][0] for x in FORMAT['GT'][s][1:] if x is not NA) for s in SAMPLES)`

tedil:
An easy fix would hence be to skip the NAs
> `count_het := sum(any(x != FORMAT['GT'][s][0] for x in FORMAT['GT'][s][1:] if x is not NA) for s in SAMPLES)`

... well, we need to replace `FORMAT['GT'][s][0]` with `next(f for f in FORMAT['GT'][s] if f is not NA)` to get the first non NA value which can be used for comparison.

dlaehnemann:
Shouldn't `count_var` be 5, as `./1` and `1/.` both contain an alternative allele? But shouldn't the current `any(gt != 0)` do exactly this?

Also, wouldn't `count_hom_ref()` be clearer than `count_ref()`? If you understand `count_ref()` not as the opposite of `count_var()`, but in a similar vein, you could expect it to count every sample that contains at least one `0` allele, including `het` sites like `0/1`.

tedil:
> Shouldn't `count_var` be 5, as `./1` and `1/.` both contain an alternative allele? But shouldn't the current `any(gt != 0)` do exactly this?

Yes! Great, less work! :wink: 

> Also, wouldn't `count_hom_ref()` be clearer than `count_ref()`? If you understand `count_ref()` not as the opposite of `count_var()`, but in a similar vein, you could expect it to count every sample that contains at least one `0` allele, including `het` sites like `0/1`.

I copied the names from SnpSift. Here's what SnpSift does with our example:
- `countHom() == 3` (+ `./.`)
- `countHet() == 6` (+ `./0`, `./1`, `1/.`, `0/.`)
- `countRef() == 4` (+ `./.`, `./0`, `0/.`)
- `countVariant() == 5` (same)

tedil:
- `count_hom`: count `x/x/…` with `x != .` cases :heavy_check_mark: (we already do that)
- `count_het`: count `x_0/x_1/...` with `x_i != x_j, i != j` cases :question: (is `.` different from `0` or `1` or `.`?)
- `count_ref`: count samples with any definitive (i.e. not `.`) allele that is `0` :x: 
- `count_var`: count samples with any definitive (i.e. not `.`) allele that is *not* `0` :heavy_check_mark:

dlaehnemann:
Well, SnpSift seems to flexibly match `.`. Mostly it is equal to `0`, but for `countHet()` it also seems to be equal to `1` -- or at least different from `0`. I would argue, that it truely is a `don't know`. So it shouldn't count as the same as anything (`0`, `1`, or `.`), but also not as clearly different from any of them. So maybe we should introduce a new return value for `nobody knows`.

Some examples:
* `0/./.` -> the missing alleles could result in the site being hom ref or het, so we genuinely `don't know`, but it could count towards a `count_any_ref_alleles_present()` function
* `0/0/.` -> same here
* `0/1/.` -> well, this clearly is het already and cannot be hom, so we can give a clear answer (and we can definitely count this towards `count_var()`, and also towards a `count_any_ref_alleles_present()` function if this is implemented)

dlaehnemann:
Basically, as soon as we have at least one missing data allele, we cannot call a site homozygous any more. But heterozygous works just fine.

tedil:
> Basically, as soon as we have at least one missing data allele, we cannot call a site homozygous any more. But heterozygous works just fine.

Great, we agree.
(But I disagree on the `nobody knows` value, we already have the `NoValue` which basically behaves like that)

dlaehnemann:
> Here's the semantics I think are the least-surprising, under the assumption that `.` is unknown data:
> 
>     * `count_hom() == 2 `: count `x/x/…` with `x != .` cases
> 
>     * `count_het() == 2`: count `x_0/x_1/...` with `x_i != x_j, i != j` (skip `.` since we can't know if it's different from `0` or `1` or `.`)
> 
>     * `count_ref() == 5 `: count samples with any definitive (i.e. not `.`) allele that is `0`
> 
>     * `count_var() == 5 `: count samples with any definitive (i.e. not `.`) allele that is _not_ `0`

Jep, this looks perfectly fine. And a `NoValue` return value sounds good to me. I would add two more functions, if only to disambiguate them clearly from the existing `count_hom()` and `count_ref()`:
* `count_hom_ref() == 1`
* `count_hom_var() == 1`

And generally, I would say this calls for lots of test cases... :p

PR #88, opened 2020-11-06
Add --version option

tedil:
This PR adds a `--version` (and the short `-v`) to the vembrane *base* parser, i.e. `vembrane --version` will output `vembrane {version}`; `vembrane filter|table --version` will *not* work.

### DIFF
--- a/vembrane/cli.py
+++ b/vembrane/cli.py
@@ -7,7 +7,7 @@ from . import __version__
 def main():
     parser = argparse.ArgumentParser()
     parser.add_argument(
-        "--version", "-v", action="version", version=f"%(prog)s {__version__}"
+        "--version", action="version", version=f"%(prog)s {__version__}"
     )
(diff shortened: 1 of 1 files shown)


vembrane/cli.py
```diff
@@ -1,10 +1,14 @@
 import argparse
 
 from .modules import filter, table
+from . import __version__
 
 
 def main():
     parser = argparse.ArgumentParser()
+    parser.add_argument(
+        "--version", "-v", action="version", version=f"%(prog)s {__version__}"
```
mbargull, line 10:
```suggestion
        "--version", action="version", version=f"%(prog)s {__version__}"
```
`-v` is also commonly used as a short option for `--verbose`. I would leave that one out.

tedil: Ah, yes, I forgot.

mbargull (approved):

PR #91, opened 2021-01-15
Add option for auxiliary files containing sets of symbols

tedil:
This PR adds the CLI option `--aux/-a NAME PATH` for reading auxiliary files, such as files containing sets of symbols which are to be filtered against. The contents of the files, separated by whitespace, are stored in a `set` and can be accessed using `AUX["NAME"]`.

Example:
```
vembrane filter '(ANN["SYMBOL"] or None) in AUX["foo"]' HG001_annotated_vep.bcf --aux foo test.txt
```
will read symbols separated by ~~whitespace~~ newlines from `test.txt` (so it does ~~not~~ matter whether symbols are on one or multiple lines) and then do a membership test.


vembrane/modules/filter.py
```diff
@@ -184,7 +194,17 @@ def statistics(
         yaml.dump(dict(counter), out)
 
 
+def read_auxiliary(aux) -> Dict[str, Set[str]]:
+    # read auxiliary files, split at any whitespace and store contents in a set
+    def read_set(path: str) -> Set[str]:
+        with open(path, "rt") as f:
+            return set(f.read().split())
```
johanneskoester, line 201: This seems dangerous to me. I think I'd rather expect each entry to be in one line. This way, there can be whitespace in the entries.

johanneskoester (changes requested):

vembrane/modules/filter.py
tedil:
```suggestion
            return set(l.strip() for l in f.readlines())
```

johanneskoester:
```suggestion
            return set(l.rstrip() for l in f)
```

johanneskoester (approved):

PR #93, opened 2021-02-04
Better handling of Number=1 and NoValue

tedil:


tedil:
This PR has evolved a bit. Now, if a field specifies `Number=1` in the header, but the actual number of values is *not* 1, a `NotExactlyOneValue` error is raised, and only in the case `Number=1` (well and `Number=A`) and `len(values) == 1` the value is unpacked.

We do not check `Number={N}` and `len(values) == N`, yet.

vembrane/modules/filter.py
```diff
@@ -16,6 +16,14 @@
 from ..representations import Environment
 from .. import __version__
 
+from argparse import Action
+
+
+class StoreMapping(Action):
+    def __call__(self, parser, namespace, values, option_string=None):
+        d = dict(m.split("=") for m in values.split(","))
+        setattr(namespace, self.dest, d)
```
mbargull, line 25:
```suggestion
        items = getattr(namespace, self.dest, None)
        if items is self.default or items is None:
            items = {}
        items.update(dict(m.split("=") for m in values.split(",")))
        setattr(namespace, self.dest, items)
```
In case you want to
- add a `default` value at some point (the logic above would discard the `default` content completely when the user provided anything),
- support multiple arguments, e.g., `vembrane --overwrite-number=X=. --overwrite-number=Y=. ...` -> `{"X": ".", "Y": "."}`.

mbargull: Also, if you want to go by "also I guess you hopefully won't have many things to overwrite...", you could also omit the `,`-splitting logic for now -- since it's a relaxation, it could be added later at any point in time, if it does seem convenient to include.

vembrane/ann_types.py
```diff
@@ -64,23 +64,27 @@ def __repr__(self):
...
 
 
-def type_info(value, number="."):
+def type_info(value, number=".", field=None, record_idx=None):
     if value is None:
         return NA
+    if number == "A":
+        if len(value) != 1:
+            raise MoreThanOneAltAllele()
+        return value[0] or NA
```
mbargull, line 73: Can any of these be empty strings at some point?
If yes, you may want to retain `value[0] if value[0] is not None else NA`.

tests/test_vembrane.py
```diff
@@ -42,8 +42,15 @@ def test_filter(testcase):
...
 
     vcf = VariantFile(path.joinpath("test.vcf"))
+
     if "raises" in config:
-        exception = getattr(errors, config["raises"])
+        exc = config["raises"]
+        try:
+            exception = getattr(errors, exc)
+        except AttributeError:
+            import builtins
```
mbargull, line 51: It's fine to import `builtins` unconditionally on the toplevel.

vembrane/ann_types.py
```diff
@@ -3,10 +3,10 @@
 from typing import Union, Iterable, Tuple, Dict, Callable, Any
 import re
 
-from .errors import MoreThanOneAltAllele
+from .errors import MoreThanOneAltAllele, NotExactlyOneValue
 
 
-class NoValue:
+class NoValue(str):
```
mbargull, line 9: Can you elaborate on this change?

vembrane/ann_types.py
tedil: Good catch. Forgot about that.

tests/test_vembrane.py
tedil: But it's also fine where it is at the moment, right? :D

vembrane/ann_types.py
tedil: If NoValue inherits from str, `re.search("something", NoValue())` does not error but just come up empty handed, which is convenient behaviour. This way, we do not have to special case / monkey patch / wrap the regex module.

tests/test_vembrane.py
mbargull: Sure ;D.
(I usually just put inconspicuous imports like `sys`/`os`/`builtins` up top and make slow-to-import or side effect causing ones conditional ;).)

vembrane/modules/filter.py
tedil: Thanks for the update logic :)

vembrane/ann_types.py
mbargull: I see, thanks.
We do have to come up with another solution though, since now, e.g., `NA == ''`.

tedil: > I see, thanks.
> We do have to come up with another solution though, since now, e.g., `NA == ''`.

No, `NA == x` is `False` for any `x`.

mbargull: Right, but with the proposed change and due to inheritance we now not only have `isinstance(NA, str) is True` but also `NoValue.__call__ is str.__call__` and `NoValue() == str()`, i.e., `NA.__eq__('')`.

mbargull: (`str.__eq__` is also inherited.)

tedil: > Right, but with the proposed change and due to inheritance we now not only have `isinstance(NA, str) is True` but also `NoValue.__call__ is str.__call__` and `NoValue() == str()`, i.e., `NA.__eq__('')`.

Not quite:
 `isinstance(NA, str) == True`
`( NoValue.__call__ is str.__call__ ) == False`
`( NoValue() == str() ) == False`
`NA.__eq__('') == False`

tedil: > (`str.__eq__` is also inherited.)

... but overloaded by

```python
    def __eq__(self, other):
        return False
```

mbargull: Ahhhhhhhhhhhhh!
Collapsed diffs in GitHub.....
Nvm

mbargull: > If NoValue inherits from str, `re.search("something", NoValue())` does not error but just comes up empty handed, which is convenient behaviour. This way, we do not have to special case / monkey patch / wrap the regex module.

Can you add this as a comment to record the intention for posterity? :)

tedil: Will do, will also update the README

vembrane/representations.py
```diff
@@ -253,7 +261,7 @@ def _get_alt(self) -> str:
         return alts[0]
 
     def _get_qual(self) -> float:
-        value = type_info(self.record.qual)
+        value = self.record.qual or NA
```
mbargull, line 264: A score of `0` could be -- although semantically questionable -- a valid input, right?

tedil: You're right; I just forgot to adapt this one.
```suggestion
        value = NA if self.record.qual is None else self.record.qual
```

mbargull: This just came to mind: Does `record.qual` come directly from Pysam as a `float`? If so, do we need or want to handle (=convert to `NA`)`float('nan')`?

mbargull:
```suggestion
        value = NA if (self.record.qual is None or isnan(self.record.qual)) else self.record.qual
```
needs an `from math import isnan` up top.
But it we could also leave it as `float('nan')` -- I haven't thought this one through ;).

tedil: Yes, pysam has `QUAL` as float (I think). However, since `QUAL` may also be missing (i.e. `.` dot), `nan` and "missing" are two different things. So I'd rather not have `nan` be `NA`, only `.` is `NA`.

mbargull: SGTM!

mbargull (approved):
nan nan nan nan nan nan nan NA Batman!

PR #95, opened 2021-03-04
fix and test table args handling

dlaehnemann:
The updates to `test_vembrane.py` feel a bit hacky, but I didn't know how to easily get `get_header()` to work with current `config.yaml` test setup. Please suggest code improvements if you can think of any!


dlaehnemann:
The failing tests are because we now test the `vembrane table` `get_header()` code. And while [in `python 3.8`, AST nodes have the `end_col_offset` attribute](https://docs.python.org/3.8/library/ast.html#ast.AST.end_col_offset), this attribute [did not exist in `python 3.7`](https://docs.python.org/3.7/library/ast.html#ast.AST.col_offset). So we can:

1. Stop testing on `python 3.7` and update the version dependency in the bioconda recipe.
2. Switch to using a library that also provides this for older python versions, e.g. [`asttokens` mentioned in this StackOverflow question](https://stackoverflow.com/a/41073256).

vembrane/modules/table.py
```diff
@@ -86,10 +86,8 @@ def tableize_vcf(
             yield env.table()
 
 
-def get_header(args) -> Optional[List[str]]:
-    if args.header == "none":
-        return
-    elif args.header == "auto":
```
tedil, line 92: What's the reasoning behind going from returning `None` to returning `"none"`?

dlaehnemann: It's actually a switch to not handle "none" in `get_header()`, but outside of it. Returning `None` here tripped up the csv.writer in this statement:
```
writer.writerow(get_header(args))
```
So I think it makes more sense to only trigger the header line writing statement when `args.header != "none"`.

dlaehnemann: Also, the previous test to do the header write  (`if args.header:`) simply returns `True` for the input string `"none"`. So this had to be changed anyways. And then the `get_header()` function doesn't have to handle the `"none"` case any more.

tedil: Ah, I see, very good.

tedil:
> The failing tests are because we now test the `vembrane table` `get_header()` code. And while [in `python 3.8`, AST nodes have the `end_col_offset` attribute](https://docs.python.org/3.8/library/ast.html#ast.AST.end_col_offset), this attribute [did not exist in `python 3.7`](https://docs.python.org/3.7/library/ast.html#ast.AST.col_offset). So we can:
> 
> 1. Stop testing on `python 3.7` and update the version dependency in the bioconda recipe.

I usually like dropping support for old/transit versions, but there wasn't much >3.7 adoption last time I checked.

> 2. Switch to using a library that also provides this for older python versions, e.g. [`asttokens` mentioned in this StackOverflow question](https://stackoverflow.com/a/41073256).

That''d be an additional dependency we don't really need, especially since this is (I think) the only place we make use of the AST at the moment. Perhaps @christopher-schroeder can chime in?

dlaehnemann:
> I usually like dropping support for old/transit versions, but there wasn't much >3.7 adoption last time I checked.
But does that really matter, as long as our favored installation route is bioconda and we can ensure `python >= 3.8` there?

> That''d be an additional dependency we don't really need, especially since this is (I think) the only place we make use of the AST at the moment. Perhaps @christopher-schroeder can chime in?
Yeah, that's probably over the top. But [the accepted answer to the above mentioned Stackoverflow question isn't encouraging](https://stackoverflow.com/a/17478618). Thus, I think the only other alternative would then be to come up with a non-AST reliant version to get the header.
BTW, I wasn't sure about the respective if-else clause anyway:
```
def get_header(args) -> List[str]:
    if args.header == "auto":
        header = args.expression
    else:
        header = args.header
    if "," not in header:
        return [header]
    else:
        # use the nodes of the first layer of the ast tree as header names
        elts = ast.parse(header).body[0].value.elts
        return [header[n.col_offset : n.end_col_offset] for n in elts]
```
I would expect a `,` in either `args.expression` or `args.header` to indicate, that we can simply split `header` upon comma to yield the `List[str]`. And the AST parsing to actually only be necessary if we have some more complicated `args.expression`. But maybe I'm overlooking something, here? I guess @christopher-schroeder has a better overview of the possible cases that can arise?

christopher-schroeder:
I got cases where the AST solution does not work. But simple splitting the expression bei "," does not either. Simple real-life Example:
`
vembrane table "CHROM, math.pow(QUAL, 10)"
`
The AST was the best thing i came up with, but as I said, it does not always work and we have to replace it anyway.

christopher-schroeder:
maybe simply counting open and closing brackets is possible and only splitting if you are on the top level is possible, but I'm not 100% sure

dlaehnemann:
Sounds like we should come up with some more test cases before going for a new solution.

dlaehnemann:
There are a number of new tests, now, and the original bugs have been removed.

In the meantime we (@tedil, @johanneskoester and I) had long discussions (elsewhere, should have done that here...) about a related issue: automatic header generation in `vembrane table` and flattening of sample-specific (`FORMAT`) annotations. Because classically, using a list comprehension over all `SAMPLES` would yield one column with the list comprehension as the header and a list entry for each record / variant site. While this might be desired in some cases, the more common use case is to generate one column per sample with only the sample-specific expression in the header. We decided to create a special function for this case, usable in the commandline expression. It is supposed to use regular Python code, so the current solution is a syntax like this, where a lambda expression using the variable sample is required:
```
for_each_sample( lambda sample: FORMAT['AD'][sample] / FORMAT['DP'][sample], colname={sample}/relative_AD)
```

Here, `colname` should be optional (and should use sample as well, when present) and if not present, this should default to the lambda expression, i.e. here it would give:
```
FORMAT['AD']['S1'] / FORMAT['DP']['S1']    FORMAT['AD']['S2'] / FORMAT['DP']['S2']
```

tedil:
Okay, so instead of the `colname="{s}/foo"` kwarg in the expression we make use of the `--header` option which vermbrane already has with a slightly modified version of `for_each_sample` which actually evaluates the body of the lambda expression.
E.g. `--header "for_each_sample(lambda s: ''.join(map(lambda x: x*2, [f'{s}/AD']))), chrom"` is valid and results in something akin to:
 ```
 HG001/ADHG001/AD	INTEGRATION/ADINTEGRATION/AD	HG003/ADHG003/AD	chrom
 ```

dlaehnemann:
I added a test case for `for_each_sample()` and using it, I fixed some things in the code it tests. However, it now complains about not knowing the name `FORMAT` in the `eval()` that tries to expand the lambda expression argument in the specified expression.

Dunno how to fix this, but I guess the eval would need to be restricted to the lambda expression argument?

tedil:
> I added a test case for `for_each_sample()` and using it, I fixed some things in the code it tests. However, it now complains about not knowing the name `FORMAT` in the `eval()` that tries to expand the lambda expression argument in the specified expression.
> 
> Dunno how to fix this, but I guess the eval would need to be restricted to the lambda expression argument?

That was just an oversight on my part; I forgot there is the `args.header == 'auto'` case -- in that case, we have to use the "replace the lambda argument name with the actual sample names" approach, instead of the "evaluate the lambda expression to string" approach (used when specifying the header manually via `--header "..."`)

dlaehnemann:
"Just" wanted to add a test case for `for_each_sample()` in the `--header` command-line argument of the `table` subcommand and then figured we should test for equal column numbers in the `--header` and main `table` expression.

And then, well, setting up tests for that I would have had to generate even more parallel code in `test_vembrane.py`, so I set out on a little detour journey to make the tests use the `execute` commands of the `table` and `filter` modules and make use of their `argparse` interfaces. This can now make systematic use of the `config.yaml` testcase specifications. And can test vcf headers, as well. All the testcases are updated and corrected accordingly.

Only two test cases still fail, both based on the same input `test.vcf` file (see [this CI run](https://github.com/vembrane/vembrane/pull/95/checks?check_run_id=2196969900)). The trouble I see is in `INFO` fields with multiple entries, and I have no clue why these things are happening:
- in `FLOAT` fields with more than one entry, some float values are truncated:
  -  for `Number=2` and `Number=R`, `1.0` is truncated to `1` (`0.5` stays untouched)
  -  for `Number=.`, `0.5` and `0.75` are both truncated to `0`
- in `STRING` fields with an arbitrary number of entries (`Number=.`), each string is simply replaced by `.`

Could these mis-formats be happening in vembrane code, or is it more likely to be happening in pysam code? Also, why wasn't this caught by the tests the way they were run before this overhaul?

tedil:
> "Just" wanted to add a test case for `for_each_sample()` in the `--header` command-line argument of the `table` subcommand and then figured we should test for equal column numbers in the `--header` and main `table` expression.
> 
> And then, well, setting up tests for that I would have had to generate even more parallel code in `test_vembrane.py`, so I set out on a little detour journey to make the tests use the `execute` commands of the `table` and `filter` modules and make use of their `argparse` interfaces. This can now make systematic use of the `config.yaml` testcase specifications. And can test vcf headers, as well. All the testcases are updated and corrected accordingly.

Great work, thanks! I did have to change the line by line comparison for `filter` to a comparison between pysam.VariantFile, though, because of the points mentioned below. For the header comparison, we still read in lines starting with "#" for now, but we should really be comparing `vcf.header.records` instead, especially considering the fact that we might also introduce bcf files.

> Only two test cases still fail, both based on the same input `test.vcf` file (see [this CI run](https://github.com/vembrane/vembrane/pull/95/checks?check_run_id=2196969900)). The trouble I see is in `INFO` fields with multiple entries, and I have no clue why these things are happening:
> 
>     * in `FLOAT` fields with more than one entry, some float values are truncated:
>       
>       * for `Number=2` and `Number=R`, `1.0` is truncated to `1` (`0.5` stays untouched)
>       * for `Number=.`, `0.5` and `0.75` are both truncated to `0`
> 
>     * in `STRING` fields with an arbitrary number of entries (`Number=.`), each string is simply replaced by `.`
> 
> 
> Could these mis-formats be happening in vembrane code, or is it more likely to be happening in pysam code? Also, why wasn't this caught by the tests the way they were run before this overhaul?

This actually happened, because the `Type=...` annotations for the respective INFO fields were incorrect, i.e. the fields FLOATX and STRINGX both had `Type=Integer` specified. I'm not sure why htslib does not error on that but just silently truncates: in the case of float to int, in the case of string to ".".

dlaehnemann:
Very nice! Appreciate all the test renames and the proper comparison of VCF files. As this should also enable testing BCFs, maybe we could add a BCF or two to the test cases?

dlaehnemann:
Or we just merge this and leave this to the open issue #97 .

tedil (approved):
LGTM

dlaehnemann:
OK, I'll have a last read through our changes and merge if I agree... ;)

dlaehnemann:
I did some final mini changes. Feel free to change further or undo stuff. And feel free to merge -- I can't write an approving review as the original author of the PR...

PR #98, opened 2021-09-21
Annotate

christopher-schroeder:
its not finished, still needs a lot of polishing and probably fixes to decrease execution time


vembrane/modules/annotate.py
```diff
@@ -0,0 +1,129 @@
...
+    with VariantFile(args.vcf) as vcf:
+        # add new info
+        for value in config["annotation"]["values"]:
+            value = value["value"]
+            vcf.header.add_meta(
+                "INFO",
+                items=[
+                    ("ID", value["vcf_name"]),
+                    ("Number", value["number"]),
+                    ("Type", "Float"),
```
tedil, line 115: It shouldn't always be float, right?

vembrane/modules/annotate.py
```diff
@@ -0,0 +1,129 @@
...
+    current_chrom = None
+    current_index = None
+    current_data = None
+    indices = []
+
+    record: VariantRecord
+    for idx, record in enumerate(vcf):
+        if not current_chrom == record.chrom:
+            current_chrom = record.chrom
+            chrom = "chr" + record.chrom
```
tedil, line 52: CHROM may or may not start with "chr" depending on the reference used / contigs defined etc. I vaguely remember reading something about "alternative names" for contig in either the SAM or the VCF specification, we should probably leverage those.

vembrane/modules/annotate.py
```diff
@@ -0,0 +1,129 @@
...
+
+        # copy only overlapping intervals
+        valid_indices = []
+        for index in indices:
+            if current_data[index]["chromEnd"] > record.start:
+                valid_indices.append(index)
+        indices = valid_indices
+
+        if len(indices):
+            pass
```
tedil, line 74: why `pass` here?

tedil (changes requested):

vembrane/modules/annotate.py
christopher-schroeder: Right, this should be extracted from the config

vembrane/modules/annotate.py
christopher-schroeder: there should be a parameter "add_chrom_to_bed" and "remove_chrom_from_bed", or better detect it automatically

vembrane/modules/annotate.py
christopher-schroeder: oh, a relict from previous versions

vembrane/modules/annotate.py
```diff
@@ -0,0 +1,129 @@
+import contextlib
+import csv
+import sys
+import yaml
+import pandas as pd
```
tedil, line 5: make sure to include newly added dependencies to the pyproject.toml, such as pandas!

vembrane/modules/annotate.py
christopher-schroeder: i implemented some kind of automatic detection

vembrane/modules/annotate.py
```diff
@@ -0,0 +1,142 @@
...
+            if len(indices):
+                env.update_data(current_data[(np.array(indices))])
+                env.update_from_record(idx, record)
+                ann_values = env.table()
+
+                for name, value in zip(
+                    map(lambda x: x["value"]["vcf_name"], config["annotation"]["values"]),
+                    ann_values,
+                ):
+                    record.info[name] = float(value)
```
tedil: We need to consider two things here:

1. `value` is not always meant to be a float, but depends on the "type" specified in the config file
2. multiple values must not overwrite existing values, but rather be appended to some collection, possibly depending on the "number" specified in the config file

vembrane/modules/annotate.py
```diff
@@ -0,0 +1,142 @@
...
+            # find the correct chrom name
+            tmp = current_chrom
+            if tmp.lower().startswith("chr"):
+                tmp = tmp[3:]
+            for prefix in ["", "chr", "Chr", "CHR"]:
+                if prefix + tmp in ann_data:
+                    chrom = prefix + tmp
+                    current_data = ann_data[chrom].to_records()
+                    current_index = 0
+                    indices = []
```
tedil, line 66: not really a big issue, but we might as well break after we find the first hit

tedil (changes requested):

vembrane/modules/annotate.py
christopher-schroeder: and removed

christopher-schroeder:
*angry* whats wrong with macos and pysam?!

vembrane/modules/annotate.py
```diff
@@ -0,0 +1,177 @@
...
+
+        yield record
+
+
+def execute(args):
+    with open(args.config, "r") as stream:
+        try:
+            config = yaml.safe_load(stream)
+        except yaml.YAMLError as exc:
+            print(exc)
```
tedil, line 129: We should re-raise that error, no?

christopher-schroeder: yes probably

vembrane/modules/annotate.py
```diff
@@ -0,0 +1,186 @@
...
+        )
+
+        # tree = IntervalTree.from_tuples(interval_tuples))
+
+    current_chrom = None
+    current_ann_data = None
+
+    record: VariantRecord
+    for idx, record in enumerate(vcf):
+        if not current_chrom == record.chrom:
```
tedil, line 82:
```suggestion
        if current_chrom != record.chrom:
```

PR #101, opened 2021-09-29
Fix subsequent ALT accesses yielding list not str

mbargull:
Reported by @christopher-schroeder.
A filter expression like `ALT == "C" and ALT == "C"` for an entry with ALT allele `C` resulted an evaluated expression `"C" == "C" and ["C"] == "C"` (i.e., erroneously yielding `False`) because we cached the whole ALT list instead of just the first value.


tedil (approved):
Thank you yet again ;)

PR #102, opened 2022-03-21
Add more annotation types

dlaehnemann:
Adds REVEL, ExACpLI and GO to `KNOWN_ANN_TYPE_MAP_VEP`, including descriptions in `Types.md`. Turns `MANE` entry into separate entries for `MANE_SELECT` and `MANE_PLUS_CLINICAL`.


tedil:
After a short discussion off of github, we noticed the following things:
VEP and/or its databases change every now and then, and not everything is versioned, i.e. there's not always a meta info entry in the VCF header for the versions, especially not for VEP plugins + databases.
The means that we have to leave the "old" `MANE` entry in the dict of annotation entries for backwards compatibility, since files using that entry exist out there in the wild.
In the future, it would be nice to have different annotation type dictionaries for different versions of annotation tools and plugins.

tedil (approved):

PR #103, opened 2022-05-16
fix: add test case to cover ANN['EXON'] fields with a range like `4-6/27`

dlaehnemann:
In the wild, I have encountered cases where the `ANN['EXON']` field is not simply `int/int`, but `int-int/int` with a range for the exon position. I start this PR with a test case to pinpoint the problem, but will try to include a fix down the line.


dlaehnemann:
This PR now makes parsing `ANN['EXON']` fields more complex, by representing exon position / number as a range, but reduces printing back to a number, if that range is of length 1. Not sure if this is the best design decision, so please feel free to suggest something better.

vembrane/ann_types.py
```diff
@@ -210,6 +211,39 @@ def typefunc(x):
...
+    @classmethod
+    def from_vep_string(cls, value: str) -> "RangeTotal":
+        v = value.split("/")
+        r = [int(s) for s in v[0].split("-")]
+        if len(r) == 1:
+            return cls(range(r[0], r[0] + 1), int(v[1]))
+        elif len(r) == 2:
+            return cls(range(r[0], r[1] + 1), int(v[1]))
+        else:
+            raise UnexpectedException(
```
tedil, line 228: `UnexpectedException` is a bit unspecific and ambiguous (and this error is not related to doctests anyways!).
It's rather a value (or parse) error I think. Not sure what to call that exactly, perhaps just use ValueError.

vembrane/ann_types.py
```diff
@@ -210,6 +211,39 @@ def typefunc(x):
...
+            )
+
+    def __str__(self):
+        if len(self.range) == 1:
+            return f"number / total: {self.range.start} / {self.total}"
+        else:
+            return f"range / total: {self.range.start} - {self.range.stop - 1} / {self.total}"
+
+    def __repr__(self):
+        return f"{self.__class__.__name__}(range={self.range.start!r}-{self.range.stop!r}, total={self.total!r})"
```
tedil, line 239:
```suggestion
return f"{self.__class__.__name__}(range=range({self.range.start!r}, {self.range.stop!r}), total={self.total!r})"
```

vembrane/ann_types.py
```diff
@@ -414,8 +448,8 @@ def convert(self, key: str, value: str) -> Tuple[str, AnnotationType]:
     "INTRON": AnnotationNumberTotalEntry(
         "INTRON", description="The intron number (out of total number)"
     ),
-    "EXON": AnnotationNumberTotalEntry(
-        "EXON", description="The exon number (out of total number)"
+    "EXON": AnnotationRangeTotalEntry(
+        "EXON", description="The exon position range (out of total number of exons)"
```
tedil, line 452: I don't quite understand if this range really gives a location estimate in bp or if it gives a range of (enumerated) exons, so perhaps we should use "index" instead of "position" if the latter is the case?
```suggestion
"EXON", description="The exon index range (out of total number of exons)"
```

Types.md
```diff
@@ -24,7 +24,7 @@ Annotations with custom types:
 |`CELL_TYPE`|`CELL_TYPE`|`List[str]`|List of cell types and classifications for regulatory feature| |
 |`CANONICAL`|`CANONICAL`|`bool`|A flag indicating if the transcript is denoted as the canonical transcript for this gene| |
 |`INTRON`|`INTRON`|`NumberTotal` with properties `number` and `total`|The intron number (out of total number)|`'ANN["INTRON"].number == 2'`|
-|`EXON`|`EXON`|`NumberTotal` with properties `number` and `total`|The exon number (out of total number)|`'ANN["EXON"].total >= 2'`|
+|`EXON`|`EXON`|`RangeTotal` with properties `range` and `total`|The known range of the exon position (out of the total number of exons)|`'ANN["EXON"].total >= 2'`|
```
tedil, line 27: See below (I somehow started with the bottom most file during review...)
```suggestion
|`EXON`|`EXON`|`RangeTotal` with properties `range` and `total`|The known range of the exon indices (out of the total number of exons)|`'ANN["EXON"].total >= 2'`|
```

tedil (changes requested):
I have some questions regarding whether this is a range of positions or a range of indices. Also I guess the exception type doesn't really fit.
Otherwise, looks good to me.

tedil:
Hm, I forgot that this change may break existing expressions that use `ANN["EXON"].number`. That should be mentioned in the changelogs, and the version bump would have to be major.

Types.md
dlaehnemann: Jep, index / indices is clearer. Thanks!

vembrane/ann_types.py
dlaehnemann: Yeah, I just took something from the IDE suggestions that sounded good. Thanks for the extra expertise sprinkled on top! :D

vembrane/ann_types.py
```diff
@@ -1,4 +1,5 @@
 from collections import defaultdict
+from doctest import UnexpectedException
```
tedil, line 2: Not needed anymore
```suggestion
```

vembrane/ann_types.py
```diff
@@ -210,6 +210,39 @@ def typefunc(x):
...
+    def from_vep_string(cls, value: str) -> "RangeTotal":
+        v = value.split("/")
+        r = [int(s) for s in v[0].split("-")]
+        if len(r) == 1:
+            return cls(range(r[0], r[0] + 1), int(v[1]))
+        elif len(r) == 2:
+            return cls(range(r[0], r[1] + 1), int(v[1]))
+        else:
+            raise ValueError(
+                "Found more than two values separated by '-', expected only a single int, or two ints separater by '-'."
```
tedil, line 228:
```suggestion
"Found more than two values separated by '-', expected only a single int, or two ints separated by '-'."
```

tedil (approved):

PR #104, opened 2022-06-21
Add `without_na`, `replace_na` and common functions from python's `statistics` module

tedil:
Rationale behind `without_na`: Sometimes fields are incorrectly specified and have NAs even though they should not; for expressions like `sum(FORMAT['DP'][s] for s in SAMPLES) > 5` this leads to `TypeErrors`. If you have such a case, `without_na` can increase convenience: `sum(without_na(FORMAT['DP'][s] for s in SAMPLES)) > 5`. Its definition is straightforward:
```python
def without_na(values: Iterable[Any]) -> "filter":
```

Rationale behind `replace_na`: Similar to `without_na`, but sometimes there may be a reasonable value to default to, e.g.: `sum(replace_na((FORMAT["DP"][s] for s in SAMPLES), 0))`.

~~Rationale behind why we supply `mean` as a custom function instead of e.g. using numpy.mean or exposing numpy itself as `np` in the globals:~~

~~1. numpy is a large dependency~~
~~2. numpy's functions operate on numpy arrays, so in the example above, the generator expression will first have to be converted to a list comprehension (resulting in an allocation) and then is converted to a numpy array internally, and only then does the actual computation happen. This is about twice as slow as the generator version (on my machine and test dataset with 3 samples).~~

~~Examples:~~
~~- custom mean: `mean(without_na(FORMAT['DP'][s] for s in SAMPLES)) > 2.4`~~
~~- numpy mean with without_na:  `np.mean(list(without_na(FORMAT['DP'][s] for s in SAMPLES))) > 2.4`~~
~~- numpy mean without without_na  `np.mean([v for s in SAMPLES if (v:= FORMAT['DP'][s]) is not NA]) > 2.4`~~

~~However, if we did include numpy in the globals, we would instantly get access to all of its functionality (including standard deviations etc). Perhaps there are alternatives to numpy which work well for generators or short lists?~~

We're now using functions from python's `statistics` module (available from python 3.4 onwards).
This for example allows writing `mean(without_na(FORMAT['DP'][s] for s in SAMPLES)) > 2.4`

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,6 +15,9 @@ The filter expression can be any valid python expression that evaluates to `bool
  * `bool`, `chr`, `float`, `int`, `ord`, `str`
  * Any function or symbol from [`math`](https://docs.python.org/3/library/math.html)
  * Regular expressions via [`re`](https://docs.python.org/3/library/re.html)
+ * custom functions:
+   * `filter_na` (keep only values that are not `NA`)
+   *  `mean` (calculate the mean of an iterable of values)
 
 ### Available fields
 The following VCF fields can be accessed in the filter expression:
@@ -93,6 +96,8 @@ Since you may want to use the regex module to search for matches, `NA` also acts
 *Explicitly* handling missing/optional values in INFO or FORMAT fields can be done by checking for NA, e.g.: `INFO["DP"] is NA`.
 
 Handling missing/optional values in fields other than INFO or FORMAT can be done by checking for None, e.g `ID is not None`.
+
+Sometimes, multi-valued fields may contain missing values; in this case, the `filter_na` function can be convenient, for example: `mean(filter_na(FORMAT['DP'][s] for s in SAMPLES)) > 2.3`.
 
 ### Auxiliary files
 `vembrane` supports additional files, such as lists of genes or ids with the `--aux NAME path/to/file` option. The file should contain one item per line and is parsed as a set. For example `vembrane filter --aux genes genes.txt "ANN['SYMBOL'] in AUX['genes']" variants.vcf` will keep only records where the annotated symbol is in the set specified in `genes.txt`.
--- a/README.md
+++ b/README.md
@@ -17,8 +17,11 @@ The filter expression can be any valid python expression that evaluates to `bool
  * Any function from [`statistics`](https://docs.python.org/3/library/statistics.html)
  * Regular expressions via [`re`](https://docs.python.org/3/library/re.html)
  * custom functions:
-   * `filter_na` (keep only values that are not `NA`)
-   *  `mean` (calculate the mean of an iterable of values)
+   * `without_na` (keep only values that are not `NA`)
+   * genotype related:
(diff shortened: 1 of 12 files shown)


README.md
```diff
@@ -14,7 +14,15 @@ The filter expression can be any valid python expression that evaluates to `bool
  * `dict`, `list`, `set`, `tuple`
  * `bool`, `chr`, `float`, `int`, `ord`, `str`
  * Any function or symbol from [`math`](https://docs.python.org/3/library/math.html)
+ * Any function from [`statistics`](https://docs.python.org/3/library/statistics.html)
  * Regular expressions via [`re`](https://docs.python.org/3/library/re.html)
+ * custom functions:
+   * `without_na` (keep only values that are not `NA`)
+   * `replace_na` (replace values that are `NA` with some other fixed value)
```
dlaehnemann, line 21: At least for `replace_na` a small example invocation is maybe a good idea to include? I guess the other functions should be clear, but this one requires an extra argument. So, just to make it as easy as possible...

README.md
```diff
@@ -94,6 +102,8 @@ Since you may want to use the regex module to search for matches, `NA` also acts
 
 Handling missing/optional values in fields other than INFO or FORMAT can be done by checking for None, e.g `ID is not None`.
 
+Sometimes, multi-valued fields may contain missing values; in this case, the `without_na` function can be convenient, for example: `mean(without_na(FORMAT['DP'][s] for s in SAMPLES)) > 2.3`.
```
dlaehnemann, line 105: Maybe the above-requested example invocation for `replace_na` can also go here, and you only quickly mention the actual signature of `replace_na(replacement=x)` above?

vembrane/globals.py
```diff
@@ -95,6 +96,47 @@
     name: mod for name, mod in vars(math).items() if not name.startswith("__")
 }
 
+
+_statistics_whitelist = {
```
dlaehnemann, line 100: Maybe `_statistics_uselist` or `_statistics_allowlist`, instead?

Also, is this just a security feature, in case unsafe functions sneak into the `statistics` module? Because this mainly sounds like an extra maintenance burden to me... :sweat_smile:

dlaehnemann (commented):
Really nice additions, thanks!

I just have some minor suggestions.

dlaehnemann:
Ah, and one more thing: we should probably add test cases for the custom functions, and a testcase or two for some `statistics` module functions, right?

README.md
tedil: Good point

README.md
tedil: Again, good point ;)

vembrane/globals.py
tedil: Right, I'll rewrite this

tedil:
> Ah, and one more thing: we should probably add test cases for the custom functions, and a testcase or two for some `statistics` module functions, right?

I added testcases for `without_na`, `replace_na` and `mean`/`median` from statistics. (Both tests for `without_na` and `replace_na` actually use `mean` from statistics as well)

PR #107, opened 2022-07-11
Update numpy version (to address dependabot warning)

tedil:
This PR updates numpy to address a dependabot security alert. While we're at it, we also update some more dependencies.
Note that the numpy update raises the minimum required python version to 3.8.


dlaehnemann (approved):
LGTM.

Also, judging from the `poetry.lock` file, this dependency update actually reduces the number of indirect dependencies quite a bit!

PR #110, opened 2022-07-14
Change backend from `pysam` to `cyvcf2`

tedil:


EQt:
Looks like cyvcf2 needs special compile flags [on MacOS](https://github.com/brentp/cyvcf2#github-building-htslib-and-cyvcf2-from-source). [Apparently](https://stackoverflow.com/a/72217172), it should suffice to set the environment variables accordingly

EQt:
`cyvcf2` fixed the issue in the new version 0.30.16 (see https://github.com/brentp/cyvcf2/pull/246 )

EQt:
After some manual tests I found that 
```sh
vembrane filter 'FORMAT["GT"]["S6"] == (None, 0)' tests/testcases/test_infotuple_comparison/test.vcf
```
is failing and smaller than the test_infotuple_comparison expression

tedil:
> After some manual tests I found that
> 
> ```shell
> vembrane filter 'FORMAT["GT"]["S6"] == (None, 0)' tests/testcases/test_infotuple_comparison/test.vcf
> ```
> 
> is failing and smaller than the test_infotuple_comparison expression

Ah, sorry, forgot to mention: yes, both pysam and cyvcf special case `FORMAT["GT"]`, and they both do it differently. I have another branch (switchable-backends, local only atm) which special cases those special cases.

EQt:
Elaborating on this it turns out that
```sh
vembrane filter 'print(repr(FORMAT["GT"]["S6"]))' tests/testcases/test_infotuple_comparison/test.vcf
```
on this  branch prints
```
InfoTuple(values=(NoValue(), 0))
```
whereas on main branch it prints
```
InfoTuple(values=(None, 0)
```

EQt:
> Ah, sorry, forgot to mention: yes, both pysam and cyvcf special case `FORMAT["GT"]`, and they both do it differently. I have another branch (switchable-backends, local only atm) which special cases those special cases.

Is this pull request still relevant, then?

tedil:
> > Ah, sorry, forgot to mention: yes, both pysam and cyvcf special case `FORMAT["GT"]`, and they both do it differently. I have another branch (switchable-backends, local only atm) which special cases those special cases.
> 
> Is this pull request still relevant, then?

Yes and no, a lot of the changes made here will also re-appear in the new branch. I'll see that I get it in a state where I can open a PR here.

christopher-schroeder:
obsolete with v2.0

PR #111, opened 2022-08-04
Replace float constants in expressions with IEEE-754-1985 32bit float constants.

tedil:
This "solves" the following issue:
The VCF specification … specifies … 32bit IEEE-754-1985 floats. Both `pysam` and `cyvcf2` use htslib. Htslib uses 32 bit c floats.
vembrane expressions use 64bit floats (see python 3 tutorial on [floating point arithmetic](https://docs.python.org/3/tutorial/floatingpoint.html) which explains the usual issues around floating point arithmetic).

In other words: there's a type disparity between float32 in VCF fields (obtained from htslib) and float64 in the expression (parsed from strings).

From a user's perspective, this can lead to unexpected behaviour:
Given the following pseudo VCF file
```
##INFO=<ID="some_float",Type=Float,Number=1,Description="">
#CHROM	POS	ID	REF	ALT	QUAL	FILTER	INFO	FORMAT
chr1	2	.	A	C	.	PASS	some_float=0.6	.
```
and the vembrane filter expression `INFO["some_float"] > 0.6`, a user would expect the record to be discarded.
However, `0.6` can never be represented exactly, in this case `c_float(0.6) == c_float(0.6000000238418579)`, leading to the record to be *kept* instead of discarded (because `c_float(0.6).value > c_double(0.6).value`).

Of course, one should not rely on exact float thresholds anyway (i.e. does it really make a difference whether the test is `foo > 0.6` or `foo > 0.6000000238418579`?).

The solution here parses the expression to obtain its AST, visit all constants that are floats, and wraps those floats in numpy.float32, compiles the modified AST and stores that instead of the string we used to store before.
This necessitates the side-effect of exposing the numpy module in the environment.

### DIFF
--- a/vembrane/representations.py
+++ b/vembrane/representations.py
@@ -167,11 +167,9 @@ class WrapFloat32Visitor(ast.NodeTransformer):
             return node
 
         return ast.Call(
-            func=ast.Attribute(
-                value=ast.Name(
-                    id="numpy", ctx=ast.Load()
-                ),  # assumes numpy is available in the global or local namespace
-                attr="float32",
+            func=ast.Name(
+                # assumes float32 is available in the globals
+                id="float32",
                 ctx=ast.Load(),
             ),
             args=[node],
--- a/vembrane/representations.py
+++ b/vembrane/representations.py
@@ -161,6 +161,26 @@ class Annotation(NoValueDict):
 UNSET = object()
 
 
+class WrapFloat32Visitor(ast.NodeTransformer):
+    def visit_Constant(self, node):
+        if not isinstance(node.n, float):
+            return node
+
+        return ast.Call(
+            func=ast.Attribute(
+                value=ast.Name(
+                    id="numpy", ctx=ast.Load()
+                ),  # assumes numpy is available in the global or local namespace
+                attr="float32",
+                ctx=ast.Load(),
+            ),
+            args=[node],
+            keywords=[],
+            starargs=None,
+            kwargs=None,
+        )
+
+
 class Environment(dict):
     def __init__(
         self,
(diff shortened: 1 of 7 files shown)


dlaehnemann (approved):
The code looks good to me. Two small things:

1. This only adds overhead in parsing the expression provided on the command-line, so no noticeable impact on run times is expected, right?
2. Maybe add a little test case to confirm the fix and avoid regressions?

tedil:
> The code looks good to me. Two small things:
> 
>     1. This only adds overhead in parsing the expression provided on the command-line, so no noticeable impact on run times is expected, right?

Well, it basically injects `numpy.float32(some_constant)` into the expression, so there is some overhead, but that should not be noticeable.
Just realized I'll also have to think about the custom parsers which produce float values, those should probably also use `numpy.float32`.

> 
>     2. Maybe add a little test case to confirm the fix and avoid regressions?

Good point!

tedil:
I added a testcase that keeps records with `0.6 < FLOAT <= 0.9` (`c_float(0.6) == c_float(0.6000000238418579)` and `c_float(0.9) == c_float(0.8999999761581421)`), which as expected fails on the main branch.

I also changed all the annotation parsers from `float` to use `numpy.float32` instead for consistency.

tedil:
… and with recent commits, instead of calling numpy.float32 during evaluation, we now simply replace python floats with `ctypes.c_float`s, e.g. a constant `0.6` is replaced with a constant `0.6000000238418579` in the expression.

tedil:
> For the benefit of my education and for future reference, could you jot down two sentences why this is better?

Yes, of course, there are actually two reasons:

1.  Imagine the expression `"INFO['float'] < 0.6"`. In the first iteration of this PR, this was transformed to `"INFO['float'] < numpy.float32(0.6)"`. That means every time the expression was evaluated, there was also a *call*  to `numpy.float32`. Now, instead of having the call, we get `"INFO['float'] < 0.6000000238418579"` with a constant instead. It's just less work to do for the interpreter.
2. The return type of `numpy.float32` is a numpy scalar, which *mostly* behaves like a plain python number type, but I'm no expert so I'm not sure if this might not cause some issues later on or introduce some unwanted side-effects and/or capabilities. Now we just have the plain python floats (> 32bit) but with the value simply being truncated to the 32bit c_float value.

vembrane/representations.py
```diff
@@ -184,7 +194,29 @@ def __init__(
...
-        self._func = eval(f"lambda: {expression}", self, {})
+
+        func = f"lambda: {expression}"
+
+        # The VCF specification only allows 32bit floats.
+        # Comparisons such as `INFO["some_float"] > CONST` may yield unexpected results,
+        # because `some_float` is a 32bit float while `CONST` is a python 64bit float.
+        # Example: `c_float(0.6) = 0.6000000238418579 > 0.6`.
+        # We can work around this by wrapping user provided floats in numpy.float32
+        # as follows:
```
dlaehnemann, line 205: This text needs to be adjusted to the newest solution:
```suggestion
        # We can work around this by wrapping user provided floats in `ctypes.c_float`,
        # which should follow the same IEEE 754 specification as the VCF spec, as most
        # C compilers should follow this standard (https://stackoverflow.com/a/17904539):
```

dlaehnemann (approved):
Thanks for the explanation, and yes, this solution probably works best with how the VCF spec defines floats.

One small thing: the explanation in the code comments should be updated. I made a suggestion, but feel free to improve the wording and correct anything I might have misunderstood.

PR #112, opened 2022-08-04
`--overwrite-number` fixes and pysam special treatment

tedil:
This PR does several things because these are all pretty much related:

1. The `FORMAT["GT"]` field is always interpreted as a list of integers by pysam, no matter the definition in the header. (The VCF specification says that the community convention is `<ID=GT,Type=String,Number=1>`, which fits for VCF, but BCF stores the GT field quite differently). We therefore now ignore the multiplicity specified in the header (by force-setting it to `"."`).

2. Switch from key-value-with-equals notation `--overwrite-number FIELD=NUM` to key-value-notation-with-space-and-nargs=2 notation `--overwrite-number FIELD NUM` so as to have consistent key-value-pair cli definitions, similar to `--aux NAME FILE`. This change also necessitates being able to use `--overwrite-number` multiple times.

3. Keep `--overwrite-number` for `INFO` fields, add `--overwrite-number-format` for `FORMAT` fields. This is a bit unfortunate, but previously both `INFO[FIELD]` and `FORMAT[FIELD]` were overwritten if the same fieldname was used in both `INFO` and `FORMAT`. Perhaps `--overwrite-number` should be renamed to `--overwrite-number-info` for consistency (or perhaps there is a better way of handling this convenience functionality)?

### DIFF
--- a/vembrane/representations.py
+++ b/vembrane/representations.py
@@ -168,7 +168,7 @@ class Environment(dict):
         ann_key: str,
         header: VariantHeader,
         auxiliary: Dict[str, Set[str]] = {},
-        overwrite_number: Dict[str, str] = {},
+        overwrite_number: Dict[str, Dict[str, str]] = {},
     ):
         self._ann_key: str = ann_key
         self._has_ann: bool = any(
@@ -209,13 +209,18 @@ class Environment(dict):
         # an index operation.
         self._numbers = {
             kind: {
-                record.get("ID"): overwrite_number.get(record.get("ID"))
+                record.get("ID"): overwrite_number.get(kind, {}).get(record.get("ID"))
                 or record.get("Number")
                 for record in header.records
                 if record.type == kind
             }
             for kind in set(r.type for r in header.records)
         }
+
+        # always explicitly set "Number" for certain fields
+        # which get special pysam treatment:
+        # - `FORMAT["GT"]` is always parsed as a list of integer values
+        self._numbers.get("FORMAT", {})["GT"] = "."
 
         # At the moment, only INFO and FORMAT records are checked
         self._header_info_fields = self._numbers.get("INFO", dict())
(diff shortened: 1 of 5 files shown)


vembrane/representations.py
```diff
@@ -209,14 +209,19 @@ def __init__(
...
                 or record.get("Number")
                 for record in header.records
                 if record.type == kind
             }
             for kind in set(r.type for r in header.records)
         }
 
+        # always explicitly set "Number" for certain fields
+        # which get special pysam treatment:
+        # - `FORMAT["GT"]` is always parsed as a list of integer values
```
dlaehnemann, line 222: Does it make sense to reference the respective pysam code, here? For quicker reference, should this need adjustments or additions in the future?

dlaehnemann (commented):
There is a test case and all the changes look good to me. One minor question. And then we need to decide on the best wording / syntax for the `--overwrite-number` option(s). Here's my 2c:

Regarding `3.`, the only other way I can come up with that works without introducing special syntax would be to require specifying the field with the dict expression syntax, so something like:
`--overwrite_number FORMAT["DP"] G`
But this seems a bit complicated and as there are exactly the two cases of `INFO` and `FORMAT`, I think I would opt for `--overwrite-number-format` and `--overwrite-number-info` for consistency. To avoid future confusion, I would deprecate `--overwrite-number`, but it should probably throw a quick explanation to use the new, properly specified versions.

tedil:
> Regarding `3.`, the only other way I can come up with that works without introducing special syntax would be to require specifying the field with the dict expression syntax, so something like: `--overwrite_number FORMAT["DP"] G` But this seems a bit complicated and as there are exactly the two cases of `INFO` and `FORMAT`, I think I would opt for `--overwrite-number-format` and `--overwrite-number-info` for consistency. To avoid future confusion, I would deprecate `--overwrite-number`, but it should probably throw a quick explanation to use the new, properly specified versions.

I agree with you in all points.

vembrane/representations.py
tedil: Not sure where in the code this happens, but there are several issues one could mention, such as: https://github.com/pysam-developers/pysam/issues/630

dlaehnemann (approved):
LGTM!

vembrane/representations.py
dlaehnemann: I guess this cross-mention does the trick, then.

PR #114, opened 2022-08-08
Allow tidy table / long format tsv output

tedil:
This PR adds a `--tidy` flag to vembrane table which allows switching from the wide format used by the `for_each_sample` feature to a long format with an additional `SAMPLE` column (always the first column) and subsequently an additional local variable called `SAMPLE`.
Example with a 3 sample input vcf: 
```sh
vembrane table "CHROM, POS, ('DP' in FORMAT and FORMAT['DP'][SAMPLE]) or '.'" input.bcf --tidy --header "CHROM, POS, DP" | head -n 7
SAMPLE	CHROM	POS	DP
A	chr1	817186	823
B	chr1	817186	1301
C	chr1	817186	883
A	chr1	817341	584
B	chr1	817341	118
C	chr1	817341	747
```


EQt:
Definitely a useful feature. I would call the parameter `--long-format` or `--long`, though. When I first saw `--tidy` I got the idea that some values would be "cleaned" (e.g. by deduplication).

EQt:
Is it hard to write some unit tests for it?

tedil:
> Definitely a useful feature. I would call the parameter `--long-format` or `--long`, though. When I first saw `--tidy` I got the idea that some values would be "cleaned" (e.g. by deduplication).

Yes, `long` is the better choice I think.

tedil:
> Is it hard to write some unit tests for it?

No, pretty straightforward. I have include two testcases for the option now.

tests/testcases/table_sample_long/expected.tsv
```diff
@@ -0,0 +1,17 @@
+SAMPLE	CHROM	FORMAT['AD'][SAMPLE][0] / FORMAT['DP'][SAMPLE] * INFO['MQ']
+SAMPLE_1	chr18	27.54098360655738
+SAMPLE_2	chr18	27.54098360655738
+SAMPLE_1	chr18	34.285714285714285
+SAMPLE_2	chr18	34.285714285714285
+SAMPLE_1	chr18	31.666666666666668
+SAMPLE_2	chr18	31.666666666666668
```
EQt, line 7: I would prefer not to use fractional numbers in unit tests (formatting might change between Python versions)

EQt:
Apart from the fractional numbers, it looks good to me.

tests/testcases/table_sample_long/expected.tsv
tedil: Right, they now get rounded to integers.

EQt (approved):
:+1:

PR #116, opened 2022-08-10
Minor formatting/Sort imports

EQt:


EQt:
Interesting that CI complains about a line that I haven't change:
```
/opt/hostedtoolcache/Python/3.8.13/x64/bin/flake8 . --exit-zero
./vembrane/representations.py:201:89: E501 line too long (89 > 88 characters)
```

tedil:
> Interesting that CI complains about a line that I haven't change:
> 
> ```
> /opt/hostedtoolcache/Python/3.8.13/x64/bin/flake8 . --exit-zero
> ./vembrane/representations.py:201:89: E501 line too long (89 > 88 characters)
> ```

Yes, stumbled upon this as well. I think it wasn't caught because it was a change made through github's editor.

If we use isort, perhaps we should also add the corresponding pre commit hook ( https://pycqa.github.io/isort/docs/configuration/pre-commit.html )

tedil (approved):
LGTM, thanks!

PR #120, opened 2022-08-12
`--preserve-order` code restructuring, adding support for BNDs without "EVENTS" but with "MATEID"

tedil:
Previously, when preserve order was not activated, there was still a second pass through the VCF to emit BNDs at the end, which can be avoided at the cost of keeping BND records that belong together (be it either due to them forming a mate-pair or being part of a larger event) in memory.
For such groups, only at least one of them has to pass the filter expression (i.e. `any`) for the whole group to be emitted. Therefore, as soon as one of them passes, we can store that bit of information in a corresponding event instance and emit all records associated with that event; any records belonging to the same event encountered afterwards can then also safely be emitted.
In the case of mate pairs (with EVENT unspecified!), we can also delete the event from the dict once both records have been emitted.


vembrane/modules/filter.py
```diff
@@ -160,67 +166,127 @@ def filter_vcf(
...
+                    events[event_name] = event
+            elif record_has_passed:
+                yield record
+
+        if len(events) > 0:
+            # output BNDs if any are were left unprocessed
+            for event_name, event in events.items():
+                if event_name and event.passed:
+                    yield from event.emit()
     else:
```
dlaehnemann, line 251: Just for readability:

```suggestion
    # this else implies `preserve_order` is set
    else:
```

Or make it explicit in the code:
```suggestion
    elif preserve_order:
```

vembrane/modules/filter.py
```diff
@@ -160,67 +166,127 @@ def filter_vcf(
...
+
+        # There may be BND record which have neither of
+        # INFO['EVENT'], INFO['MATEID'] or even ID specified.
+        # In that case, we still need a pseudo-event name,
+        # which we construct from the record's index in the file.
+        def get_event_name(record: VariantRecord) -> str:
```
dlaehnemann, line 257: Why isn't this function also used in the above `if not preserve_order` branch? I'm probably missing a subtle difference here, but just want to make sure this is intended.

vembrane/modules/filter.py
```diff
@@ -160,67 +166,127 @@ def filter_vcf(
...
+                or f"DUMMY_ID: {idx}"
+            )
+            return event_name
+
         # If order *is* important, the first pass cannot emit any records but only
         # keep track of breakend events. The records will only be emitted during the
         # second pass
+        events: Set[str] = set()
         for idx, record in enumerate(vcf):
             is_bnd = "SVTYPE" in info_keys and record.info.get("SVTYPE", None) == "BND"
```
dlaehnemann, line 274:
```suggestion
            is_bnd = has_svtype and record.info.get("SVTYPE", None) == "BND"
```

Or turn `is_bnd` into a little helper function?

dlaehnemann (approved):
LGTM, just some minor questions. Most important is probably the one about the test case.

tests/testcases/test_breakend_constellations/expected.vcf
```diff
@@ -0,0 +1,26 @@
...
+fake	200	bnd_2_0	A	]fake:250]A	.	PASS	SVTYPE=BND;EVENT=an_event;ints=1,2	.	.
+fake	250	bnd_2_1	C	C[fake:200[	.	PASS	SVTYPE=BND;EVENT=an_event	.	.
+fake	325	.	T	G	.	PASS	ints=0,2	.	.
+fake	300	bnd_3_0	A	]fake:350]A	.	PASS	SVTYPE=BND;MATEID=bnd_3_1;EVENT=mate_pair_event;ints=1,2	.	.
+fake	350	bnd_3_1	C	C[fake:300[	.	PASS	SVTYPE=BND;MATEID=bnd_3_0;EVENT=mate_pair_event	.	.
+fake	400	bnd_4_0	A	]fake:450]A	.	PASS	SVTYPE=BND;ints=1,2	.	.
+fake	425	.	T	G	.	PASS	ints=0,2	.	.
+fake	500	.	A	]fake:550]A	.	PASS	SVTYPE=BND;ints=1,2	.	.
+fake	525	.	T	G	.	PASS	ints=0,2	.	.
+fake	321681	bnd_W	G	G]fake:421681]	.	PASS	SVTYPE=BND;MATEID=bnd_U;EVENT=INV0	.	.
```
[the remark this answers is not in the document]
tedil: What you're missing here is that all of `bnd_[UVWX]` are part of the `EVENT` named `INV0`, and `bnd_X` does have `ints=1,2` set, which is why we have to keep all of them.

tests/testcases/test_breakend_constellations/test.vcf
```diff
@@ -0,0 +1,28 @@
...
+##INFO=<ID=MATEID,Number=.,Type=String,Description="ID of mate breakends">
+##INFO=<ID=EVENT,Number=1,Type=String,Description="ID of event associated to breakend">
+#CHROM	POS	ID	REF	ALT	QUAL	FILTER	INFO	FORMAT	SAMPLE1
+fake	100	bnd_1_0	A	]fake:150]A	.	PASS	SVTYPE=BND;MATEID=bnd_1_1;ints=1,2	.	.
+fake	125	.	T	G	.	PASS	ints=0,2	.	.
+fake	150	bnd_1_1	C	C[fake:100[	.	PASS	SVTYPE=BND;MATEID=bnd_1_0	.	.
+fake	200	bnd_2_0	A	]fake:250]A	.	PASS	SVTYPE=BND;EVENT=an_event;ints=1,2	.	.
+fake	225	.	T	G	.	PASS	ints=0,2	.	.
+fake	250	bnd_2_1	C	C[fake:200[	.	PASS	SVTYPE=BND;EVENT=an_event	.	.
+fake	300	bnd_3_0	A	]fake:350]A	.	PASS	SVTYPE=BND;MATEID=bnd_3_1;EVENT=mate_pair_event;ints=1,2	.	.
```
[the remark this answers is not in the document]
tedil: Good point, will include that

vembrane/modules/filter.py
```diff
@@ -160,67 +166,127 @@ def filter_vcf(
...
+                        # … emit all records associated with it
+                        yield from event.emit()
+
+                        # in the case of a simple mate pair, we can delete the event
+                        # at this point, because no more records will be added to it
+                        if event.is_mate_pair():
+                            del events[mate_pair]
+                else:
+                    # if there's no entry for the event or mate pair yet, create one
+                    is_mate_pair = mate_pair and mate_pair == event_name
```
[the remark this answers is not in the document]
tedil: Both `mate_pair` and `event_name` can be `None`. So we check for `mate_pair  # (is not None)` here to make sure we're actually in the "this BND is part of a mate pair without an event tag specified" case.

vembrane/modules/filter.py
tedil: Good idea, done.

vembrane/modules/filter.py
tedil: There is indeed a minor difference as to how exactly records are filtered / "pass" information is memorized here:

In the case where order does not matter, we keep the records in memory until we have sufficient information.
Records are always associated to a pseudo-event, which is either defined through the EVENT tag or two MATE_IDs combined.
However, in the case neither of those are set, the information in the VCF regarding breakends is basically under-specified (it's not trivial to link the correct breakend records to form mates etc without the mate id information etc), so we treat them as regular non-BND records (which corresponds to the `None` case in `event or mate_pair or None`).

When order does matter, the way we keep track of which events to keep/discard changes slightly:
we just keep a set of event names which we populate in a first pass (i.e. if a record passes the expression and belongs to some event, that event is added to the set).
In a second pass, we check if the event name was memorized during the first pass, and if so, output that record.
In the case where both EVENT and mate-pair information are unavailable, we still need store *some* key/event/name in the `events` set, and that key is then either the record's ID or simply the index in the file.

vembrane/modules/filter.py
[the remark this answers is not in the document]
dlaehnemann: Ah OK, I just checked and this is would otherwise be a problem:
```
>>> None == None
True
```

vembrane/modules/filter.py
dlaehnemann: Thanks for the detailed info. I'm just worried, that code might diverge in future edits and lead to inconsistent behaviour. So maybe it makes sense to reduce this to a function that just returns `(event_name, mate_id)` and then have the different handling on those results?

dlaehnemann (approved):
Abgenickt. :heavy_check_mark:

vembrane/modules/filter.py
tedil: That's a good suggestion, thanks, implemented now ;)

PR #122, opened 2022-08-13
TypeVar instead of Any

EQt:
- Prefer concrete type parameter `T` to general `Any`
- Document non-private methods

### DIFF
--- a/vembrane/globals.py
+++ b/vembrane/globals.py
@@ -108,10 +108,12 @@ T = TypeVar("T")
 
 
 def without_na(values: Iterable[T]) -> Iterable[T]:
+    """Keep only values that are not `NA`."""
     return filter(lambda v: v is not NA, values)
 
 
 def replace_na(values: Iterable[T], replacement: T) -> Iterable[T]:
+    """Replace values that are `NA` with `replacement`."""
     for v in values:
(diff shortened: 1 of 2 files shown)


vembrane/globals.py
```diff
@@ -104,11 +104,16 @@
...
+
+
+def without_na(values: Iterable[T]) -> Iterable[T]:
+    """Keep only values that are not `NA`."""
     return filter(lambda v: v is not NA, values)
 
 
-def replace_na(values: Iterable[Any], replacement: Any) -> Iterable[Any]:
+def replace_na(values: Iterable[T], replacement: T) -> Iterable[T]:
+    """Replace values that are `NA` with `replacement`."""
```
tedil, line 116: Theoretically, values are `Iterable[Union[T, NoValue]]` and the replacement may even be something entirely different. That doesn't make much sense, though, so I'm fine with this as is.

tedil (approved):

vembrane/globals.py
EQt: You're right. Forget about that. Shall I fix that?

tedil: If you have the time, sure, otherwise, I think it's not too important

PR #123, opened 2022-08-15
Mention possibility of `NA` in annotation.

EQt:


vembrane/globals.py
```diff
@@ -107,12 +107,12 @@
...
 
 
-def without_na(values: Iterable[T]) -> Iterable[T]:
+def without_na(values: Iterable[Union[T, NoValue]]) -> Iterable[T]:
     """Keep only values that are not `NA`."""
     return filter(lambda v: v is not NA, values)
 
 
-def replace_na(values: Iterable[T], replacement: T) -> Iterable[T]:
+def replace_na(values: Iterable[Union[T, NoValue]], replacement: T) -> Iterable[T]:
```
tedil, line 115: Or even:
```suggestion
R = TypeVar("R")
def replace_na(values: Iterable[Union[T, NoValue]], replacement: R) -> Iterable[Union[T, R]]:
```
(because we do no checks whether the replacement is actually of the same type as the value type of the iterable.)

EQt: I don't think this is necessary, because in most cases you will want to replace `NA` by something of the same kind. And in case you want to insert something different, you could always substitute `T` with `Union[T, R]`.

By the way, `pytype` checker is okay with the suggested, shorter version.

tedil: Exactly! Good.

tedil (approved):

PR #126, opened 2022-08-18
Add `tag` subcommand

tedil:
This PR adds a `tag` subcommand which is similar to `filter`, but instead of actually filtering records, it adds corresponding `FILTER` tags to the record.

Example:
```sh
vembrane tag AT_LEAST_3_DATASETS="INFO['datasets'] >= 3",ILLUMINA="'Illumina' in INFO['platformnames']" input.vcf
```
The resulting output has the exact same records as the input, but records which pass an expression have the corresponding tag added to their `FILTER` field.

Not sure about the CLI here. At the moment, it takes a single `tag=expr[,tag=expr]*` argument; instead, it could for example be multiple `--add tag=expr` options (or `--add tag expr` to keep in-sync with the other key-value options that vembrane already has).

At the moment, the description used for the FILTER header line is simply the expression.


johanneskoester:
If `--add tag expr` matches the style of the other subcommands, I'd go for that.

johanneskoester:
Apart from that: this is a great idea, and certainly very useful for some people.

tedil:
> If `--add tag expr` matches the style of the other subcommands, I'd go for that.

That's the thing; for filter it is simply `filter expression` with one single expression *argument*; but tag always needs an association between a tag and an expression (i.e. tag=expression) but also naturally support multiple expressions.
For vembrane's *options*, it's indeed `--option key value`.

So you can either look at it as `tag [multiple-expression-arguments]` or `tag [--with-required-options-to-speficy-tag=expr]`.

I think I actually prefer to use multiple arguments with `tag [tag=expr ]+` over a single argument `tag [tag=expression,]+` and that in turn over using required options, because tags and expressions aren't really options but arguments to the `tag` subcommand.

tedil:
This now uses `--tag/-t TAG=EXPRESSION` (required) options.
Also added a section to the README with two easy examples.

FelixMoelder (approved):

vembrane/modules/tag.py
```diff
@@ -0,0 +1,211 @@
...
+        "Example: `--overwrite-number-format DP 2`",
+    )
+    parser.add_argument(
+        "--invert",
+        "-i",
+        action="store_true",
+        help="Invert/negate the expression. "
+        "By default, records for which the tag expressions pass are tagged."
+        "To tag records for which the tag expressions fail, use this option.",
+    )
```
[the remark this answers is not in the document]
dlaehnemann: Also, another option would be to offer a command-line option `--vcf-spec` to completely fulfill the VCF spec `4.4`, including setting `.` if no filter is applied (which does not make much sense for the `tag` subcommand, but hey) and setting `PASS` if all specified filters pass.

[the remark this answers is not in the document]
dlaehnemann: And I would add to the command-line help (and probably the README.md) which symbols are not allowed in tag names, namely whitespace, `,`, `;` `<` and `>`, right?

johanneskoester:
> This now uses `--tag/-t TAG=EXPRESSION` (required) options -- deviating from the `--aux NAME FILE` syntax. Hmm. Should probably both either use `K=V` or `K V` syntax.
> 
> Also added a section to the README with two easy examples.

Yes, it should be consistent. I personally find the notation with the `=` more readable. But a breaking change in --aux is also not nice. One could support both syntaxes, while documenting primarilty the one with `=`?

dlaehnemann (approved):
This generally looks very good to me!

I have some suggestions regarding `--invert` and the VCF spec. But those are nothing that absolutely needs to be changed.

vembrane/modules/tag.py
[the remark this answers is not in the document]
tedil: Just looked it up again:
> String, no whitespace or semicolons permitted, duplicate values not allowed

Seems I was too strict ;) Will modify accordingly

tedil:
> > > This now uses `--tag/-t TAG=EXPRESSION` (required) options -- deviating from the `--aux NAME FILE` syntax. Hmm. Should probably both either use `K=V` or `K V` syntax.
> > > Also added a section to the README with two easy examples.
> > 
> > 
> > Yes, it should be consistent. I personally find the notation with the `=` more readable. But a breaking change in --aux is also not nice. One could support both syntaxes, while documenting primarilty the one with `=`?
> 
> laughing I find the version with just a whitespace less cluttered and more readable. But I really like the idea of supporting both, but consistently documenting one (I'd say whichever Till prefers;).

I personally prefer the `=` variant. I'm not sure how easy it would be to support both, as I'm not particularly well versed in `argparse`, which probably defaults to splitting arguments on whitespace. In that case, the number of arguments for the two variants is different (1 for `tag=expr`, 2 for `tag expr`), so it's probably not particularly easy to support both at the same time.

tedil:
> > This now uses `--tag/-t TAG=EXPRESSION` (required) options -- deviating from the `--aux NAME FILE` syntax. Hmm. Should probably both either use `K=V` or `K V` syntax.
> > Also added a section to the README with two easy examples.
> 
> Yes, it should be consistent. I personally find the notation with the `=` more readable. But a breaking change in --aux is also not nice. One could support both syntaxes, while documenting primarilty the one with `=`?

I agree. On the other hand, versions < 1.0 aren't classified as stable anyways, so we might as well use the opportunity to change that?

tedil:
Back to `key=value` arguments, making sure to use those everywhere.
This implies breaking changes for options `--aux name file` and `--overwrite-number-* field number`.
(Also accidentally added `--aux` to the table subcommand *in this branch* instead of a new branch)

README.md
```diff
@@ -179,7 +179,44 @@ Handling missing/optional values in fields other than INFO or FORMAT can be done
...
+  ```
+* Tag records with quality at least 30 with a tag named `quality_at_least_30`
+  and records which have "Illumina" in their list of platforms with a tag named `illumina`:
+  ```sh
+  vembrane tag -t quality_at_least_30 "QUAL >= 30" -t illumina "'Illumina' in INFO['platformnames']" variants.vcf
+  ```
+* Tag records with quality *less than* 30 with a tag named `q30` using the `--invert` option:
+    ```sh
+    vembrane tag --tag-mode fail --tag q30 "QUAL >= 30" variants.vcf
+    ```
```
dlaehnemann, line 211:
```suggestion
* Tag records with quality *less than* 30 with a tag named `q_below_30` using the `--tag-mode fail` setting:
    ```sh
    vembrane tag --tag-mode fail --tag q_below_30 "QUAL >= 30" variants.vcf
    ```
```

README.md
```diff
@@ -179,7 +179,44 @@ Handling missing/optional values in fields other than INFO or FORMAT can be done
...
+* Tag records with quality *less than* 30 with a tag named `q30` by negating the expression itself:
+    * ```sh
+      vembrane tag --tag q30 "not (QUAL >= 30)" variants.vcf
+      ```
+    * ```sh
+      vembrane tag --tag q30 "QUAL < 30" variants.vcf
+      ```
```
dlaehnemann, line 218:
```suggestion
* Tag records with quality *less than* 30 with a tag named `q_below_30` by negating the expression itself:
    * ```sh
      vembrane tag --tag q_below_30 "not (QUAL >= 30)" variants.vcf
      ```
    * ```sh
      vembrane tag --tag q_below_30 "QUAL < 30" variants.vcf
      ```
```

README.md
dlaehnemann: I'd also go for descriptive tags here, even though the `q30` would correspond to the VCF spec example. But I think it's always best to encourage descriptive and informative names for "things".

vembrane/modules/tag.py
```diff
@@ -0,0 +1,220 @@
...
+    parser.add_argument(
+        "--overwrite-number-info",
+        nargs=2,
+        action="append",
+        metavar=("FIELD", "NUMBER"),
+        default=[],
+        help="Overwrite the number specification for INFO fields "
+        "given in the VCF header. "
+        "Example: `--overwrite-number cosmic_CNT .`",
```
dlaehnemann, line 75:
```suggestion
        "Example: `--overwrite-number cosmic_CNT=.`",
```

vembrane/modules/tag.py
```diff
@@ -0,0 +1,220 @@
...
+    )
+    parser.add_argument(
+        "--overwrite-number-format",
+        nargs=2,
+        action="append",
+        metavar=("FIELD", "NUMBER"),
+        default=[],
+        help="Overwrite the number specification for FORMAT fields "
+        "given in the VCF header. "
+        "Example: `--overwrite-number-format DP 2`",
```
dlaehnemann, line 85:
```suggestion
        "Example: `--overwrite-number-format DP=2`",
```

vembrane/modules/tag.py
```diff
@@ -0,0 +1,220 @@
...
+        "-m",
+        default="pass",
+        choices=["pass", "fail"],
+        help="Set, whether to tag records that pass the tag expression(s), "
+        "or records that fail them."
+        "By default, records for which the tag expression(s) pass are tagged. "
+        "This allows for descriptive tag names such as `q_at_least_30` "
+        "to correspond to `QUAL >= 30`. "
+        "However, the VCF specification (`v4.4`) defines tags to be set when a "
+        "filter expression is failed.",
```
dlaehnemann, line 98: Slightly improve my wording, maybe?
```suggestion
        help="Set, whether to tag records that pass the tag expression(s), "
        "or records that fail them."
        "By default, vembrane tags records for which the tag expression(s) pass. "
        "This allows for descriptive tag names such as `q_at_least_30`, "
        "which would correspond to an expression `QUAL >= 30`. "
        "However, the VCF specification (`v4.4`) defines tags to be set when a "
        "filter expression is failed, so vembrane also offers the `fail` mode.",
```

dlaehnemann (approved):
Still looking good. Added some more suggestions. Maybe:

1. Double-check again, that it is always `key=value` everywhere, even in help examples (I caught some of those, but maybe there are more---so maybe a bit of semi-systematic searching in the code base?).
2. Maybe another test or two would be good. I'd definitely add one example for the `--tag-mode fail`, and then maybe one where the tag is rejected by `check_tag()`? Just some more code coverage...

tedil:
> Still looking good. Added some more suggestions. Maybe:
> 
>     1. Double-check again, that it is always `key=value` everywhere, even in help examples (I caught some of those, but maybe there are more---so maybe a bit of semi-systematic searching in the code base?).
 
I think we caught all instances now.

>     2. Maybe another test or two would be good. I'd definitely add one example for the `--tag-mode fail`, and then maybe one where the tag is rejected by `check_tag()`? Just some more code coverage...

Yep, indeed. There are now two more tests for those two cases.
Thank you for the suggestions, pointers, etc!

vembrane/modules/tag.py
```diff
@@ -0,0 +1,220 @@
...
+from .filter import DeprecatedAction, read_auxiliary
+
+
+def add_subcommand(subparsers):
+    parser = subparsers.add_parser("tag")
+    parser.register("action", "deprecated", DeprecatedAction)
+    parser.add_argument(
+        "--tag",
+        "-t",
+        help="Tag records using the FILTER field.",
```
dlaehnemann, line 29:
```suggestion
        help="Tag records using the FILTER field."
        "Example: `--tag q_above_30=\"not (QUAL<=30)\"`",
```

vembrane/modules/tag.py
```diff
@@ -0,0 +1,220 @@
...
+            record, keep = test_record(env, idx, record, ann_key)
+            if invert:
+                keep = not keep
+            if keep:
+                record.filter.add(tag)
+        yield record
+
+
+def check_tag(tag: str):
+    if re.search(r"[\s;]|0", tag):
```
dlaehnemann, line 153: Why not just:
```suggestion
    if re.search(r"[\s;0]", tag):
```

dlaehnemann (commented):
Already merged, but I have two more suggestions... :grimacing:

PR #128, opened 2022-09-22
feat: support for accessing END position

johanneskoester:
Before, it could not be accessed, because pysam hides it from the INFO dict. Now, we infer it from record.stop, and expose it both via `INFO["END"]` (because users will expect that to work as well since END can be specified in an info field) and `END`.


vembrane/representations.py
```diff
@@ -101,6 +101,11 @@ def __init__(
         self._info_dict = {}
 
     def __getitem__(self, item):
+        if item == "END":
+            # pysam removes END from info. In order to fit with user expectations,
+            # (they will expect INFO["END"] to work) we emulate it being present by
+            # inferring it from pysams record representation.
+            return get_end(self._record)
```
tedil, line 108: I think this should be in the `except` arm, and the `_info_dict` should be updated accordingly, just like every other access.

README.md
```diff
@@ -62,20 +62,21 @@ The filter expression can be any valid python expression that evaluates to `bool
...
-|`FORMAT`|`Dict[str, Dict[str, Any¹]]`| `Format -> (Sample -> Value)` | `FORMAT["DP"][SAMPLES[0]] > 0` |
-|`SAMPLES`|`List[str]`| `[Sample]`  |  `"Tumor" in SAMPLES` |
-|`INDEX`|`int`| Index of variant in the file  |  `INDEX < 10` |
+| Name      | Type                         | Interpretation                     | Example expression             |
+| --------- | ---------------------------- | ---------------------------------- | ------------------------------ |
+| `INFO`    | `Dict[str, Any¹]`            | `INFO field -> Value`              | `INFO["DP"] > 0`               |
+| `ANN`     | `Dict[str, Any²]`            | `ANN field -> Value`               | `ANN["Gene_Name"] == "CDH2"`   |
+| `CHROM`   | `str`                        | Chromosome Name                    | `CHROM == "chr2"`              |
+| `POS`     | `int`                        | Chromosomal position (1-based)     | `24 < POS < 42`                |
+| `END`     | `int`                        | Chromosomal end position (1-based) | `24 < END < 42`                |
```
tedil, line 71: I'm not sold on exposing `END` in addition to `INFO["END"]` yet. Every entry in this table is related to an actual VCF column, with the exceptions of `ANN` (which is core to vembrane) and `INDEX` which just keeps track of record indices. `END` would be the first truly derived property, which is either `POS + len(REF)` or `INFO["END"]`. But at least it's a reserved INFO key in the VCF spec v4.3.

tests/testcases/test_filter_end/config.yaml
```diff
@@ -0,0 +1,2 @@
+function: "filter"
+expression: 'END == 1500 or INFO["END"] == 1400'
```
tedil, line 2: `1400 != 1500`

tedil (changes requested):
Seems like a reasonable addition, but not quite sure of the details / consequences yet.

tests/testcases/test_filter_end/config.yaml
tedil: nvm

README.md
johanneskoester: Indeed. But it makes sense to me because it would feel not natural to access INFO["END"] for a small variant where there is no END tag in the record at all.

vembrane/representations.py
johanneskoester: Good point.

README.md
tedil: Yes, that is true, I'm okay with having `END` exposed then.
What happens for breakends across contigs?

johanneskoester: For breakends, having END in INFO makes little sense. I've checked it though: pysam would use whatever value is defined there. If you don't specify END for a breakend, it will be the breakend position itself, like with an SNV.

johanneskoester: We could alternatively set END to NA in case of a breakend.

johanneskoester: I think that should be more correct. What do you think?

tests/testcases/test_filter_end_breakend/test.vcf
```diff
@@ -0,0 +1,42 @@
...
+##FORMAT=<ID=FLOATR,Number=R,Type=Float,Description="One float per allele">
+##FORMAT=<ID=STRINGR,Number=R,Type=String,Description="One string per allele">
+##FORMAT=<ID=INTG,Number=G,Type=Integer,Description="One integer per genotype">
+##FORMAT=<ID=FLOATG,Number=G,Type=Float,Description="One float per genotype">
+##FORMAT=<ID=STRINGG,Number=G,Type=String,Description="One string per genotype">
+##FORMAT=<ID=INTX,Number=.,Type=Integer,Description="A variable number of integers">
+##FORMAT=<ID=FLOATX,Number=.,Type=Float,Description="A variable number of floats">
+##FORMAT=<ID=STRINGX,Number=.,Type=String,Description="A variable number of strings">
+#CHROM	POS	ID	REF	ALT	QUAL	FILTER	INFO	FORMAT	Sample
+chr1	1	.	G	 G]17:198982]	276	PASS	FLAG;FLOAT=0.5;STRING=String;INT2=1,2;FLOAT2=0.5,1.0;STRING2=String1,String2;INTA=1;FLOATA=0.5;STRINGA=StringA;INTR=1,2;FLOATR=0.5,1.0;STRINGR=StringR1,StringR2;INTX=1,2,3,4;FLOATX=0.5,0.75;STRINGX=String1,String2,String3	GT:INT:INT2:INTA:INTR:INTG:INTX	0/1:1:1,2:1:1,2:1:1,2,3,4
```
tedil, line 42: reference sequence `17` is not defined in the header, might be problematic

tests/testcases/test_filter_end/expected.vcf
```diff
@@ -0,0 +1,44 @@
+##fileformat=VCFv4.1
+##FILTER=<ID=PASS,Description="All filters passed">
+##fileDate=20210218
+##contig=<ID=chr1,length=1>
```
tedil, line 4: chromosome length should be compatible with the positions specified in the records

tedil (changes requested):

README.md
tedil: Semantically, I think NA makes more sense, because it's difficult or impossible to know what the "end of a variant" in the context of breakends actually is.

tedil: Can you update the README accordingly?

tests/testcases/test_filter_end/expected.vcf
johanneskoester: all fixed

tests/testcases/test_filter_end_breakend/test.vcf
johanneskoester: fixed

README.md
johanneskoester: done

PR #129, opened 2022-10-01
Fix: False HeaderWrongColumnNumber error when comma in non-top-level expression

christopher-schroeder:
get top level for expression instead of split(,)


tedil (commented):
It would be nice if you could add your example as a test case, but other than that, looks good, thanks for the fix ;)

tedil (approved):

PR #130, opened 2022-10-21
Add examples for working with sample specific values and genotypes

tedil:
This PR adds examples for working with sample specific values and genotypes to the README.
Showcases:
 - sample access by both name and index
 - custom genotype related functions `is_het` and `is_hom_ref`
 - access to `FORMAT` values in general

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,7 +131,7 @@ The following VCF fields can be accessed in the filter expression:
 * Filter on sample specific values:
   ```sh
   vembrane filter \
-    'mean(FORMAT["DP"][s] for s in SAMPLES) > 10. and (FORMAT["AD"][SAMPLES[0]] 0 + FORMAT["AD"]["specific_sample_name"]) > 0' \
+    'mean(FORMAT["DP"][s] for s in SAMPLES) > 10 and (FORMAT["AD"][SAMPLES[0]] + FORMAT["AD"]["specific_sample_name"]) > 0' \
     variants.vcf
   ```
 * Filter on genotypes for specific samples (named "kid", "mom", "dad"):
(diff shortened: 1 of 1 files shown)


dlaehnemann (commented):
I think there are several very good points in here, but maybe we can reorganize a bit to make them clearer?

My suggestions would be:
1. Use the filtering on sample-specific values (maybe simplified a bit to the part after the `and`?) to show that you can reference samples both by (overall) index, by indexing into the `SAMPLES` list and by directly using a sample name. So have 3 example invocations here.
2. Give only one example of how to filter on the `GT` field.
3. Give the example of filtering using the genotype functions on specific samples.

Does this make sense?

README.md
```diff
@@ -128,6 +128,35 @@ The following VCF fields can be accessed in the filter expression:
     'not {"pathogenic", "likely_pathogenic", "drug_response"}.isdisjoint(ANN["CLIN_SIG"])' \
     variants.vcf
   ```
+* Filter on sample specific values:
+  * by sample name:
+    ```sh
+    vembrane filter 'FORMAT["AD"]["specific_sample_name"] > 0' variants.vcf
```
dlaehnemann, line 134: `FORMAT["AD"]` will have one entry per allele, right? So we should probably use a simpler entry, like the `FORMAT["DP"]`used below. Otherwise, we would need to somehow make the handling of `A` type entries clear in this context, but it is probably best explained on its own.

dlaehnemann (approved):
Just one minor comment, but generally good to go, I'd say.

README.md
tedil: Yup, thanks ;)

PR #135, opened 2023-03-15
No more custom ANN names: always use names from VEP / snpEff as is

tedil:
For some annotations, `vembrane` had custom names, for example `cDNA` for VEP's `cDNA_position` and snpEff's `cDNA.pos / cDNA.length`.
This can be quite counterintuitive, as users would expect being able to access e.g. `cDNA_position` by that name (in the case of VEP), and not by `cDNA`.

This PR also re-formats the ann_types.md such that it is easier to read for humans when not rendered to html.


dlaehnemann (approved):
The cleanup of the tables that now have a column less (which contained a lot of redundant information) also very strongly suggests, that this is the better way... :D

FelixMoelder:
Just tried and this is working now. Without custom names this is probably much clearer.

PR #136, opened 2023-03-22
Attribute access on `NA` values results in `NA`

tedil:
For custom types such as `RangeTotal` with attributes named "range" and "total", we run into issues when the types are replaced with `NA`.
Take for example the ANN["EXON"] field:
- When a string like `"3-6/8"` is given, everything is fine, since `ANN["EXON"].range == range(3, 6)` and `ANN["EXON"].total == 8`.
- When however the field is empty, we get a value of `NA` and `ANN["EXON"].range` will result in an `AttributeError`.

To handle this, we can either allow `hasattr`, such that handling such cases can be expressed similar to the following example:  `ANN["EXON"].range.start if hasattr(ANN["EXON"], "range") else NA`.
This however can quickly become unwieldy, so we chose to simply implement `__getattr__` for `NoValue` (i.e. `NA`) such that it issues a warning and returns `NA` again.

- [x] Document this behaviour


FelixMoelder (approved):

vembrane/ann_types.py
```diff
@@ -49,6 +49,13 @@ def __repr__(self):
     def __hash__(self) -> int:
         return super().__hash__()
 
+    def __getattr__(self, item):
+        print(
+            f"Warning: attribute '{item}' does not exist, returning NA instead.",
```
dlaehnemann, line 54: Maybe add more context here, using `self`? And maybe add suggestions, what might be wrong. For example:

* "Annotation '{self}' does not exist in this particular record?"
* "Maybe you mistyped the intended attribute, did you really mean '{item}'?"

tedil: `NoValue` is a general value-is-missing indicator, so this is not specific to ANN, and `self` is always `NoValue`.
To fix this only for specific ANN types, we'd have to choose a different route, not quite sure what a good solution would be.

PR #138, opened 2023-03-23
`PosRange` specific fixes

tedil:
1) Assume a single position corresponds to the end-exclusive range `position-(position + 1)`.
2) When checking if either of start or end position is unspecified, check for `NA` explicitly (not with `!=`, which is always true for `NA`!).


tedil:
I was unable to find a stringent definition of the `*_position` fields in VEP.
What we've seen in the wild are the following options:
 - a single number: `42`
 - a range of numbers: `42-47`
 - a range of numbers with
   - no known start: `?-47`
   - no known end: `42-?`

For consistency, we assume that a single number corresponds to the range `number-(number+1)`, e.g. `42` → `42-43`

PR #141, opened 2023-03-30
Fix regex for checking tag names, add more help to tag option

tedil:


dlaehnemann (approved):
And maybe also add the info on unsupported symbols in tag names in the main `README.md` intro of the `tag` subcommand? This is probably where users will first find out about the subcommand, so it would help to already have this important information right there.

PR #142, opened 2023-03-30
Support `--aux` for vembrane table.

tedil:
This PR adds support for the `--aux name=file` option also available for filter and tag.
It also removes the `--all` option, which I had no idea existed and did not do anything.

### DIFF
--- a/vembrane/modules/table.py
+++ b/vembrane/modules/table.py
@@ -2,12 +2,12 @@ import contextlib
 import csv
 import sys
 from sys import stderr
-from typing import Dict, Iterator, List, Optional
+from typing import Dict, Iterator, List, Optional, Set
 
 import asttokens
 from pysam.libcbcf import VariantFile, VariantRecord
 
-from ..common import AppendKeyValuePair, check_expression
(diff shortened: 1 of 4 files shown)


vembrane/modules/table.py
```diff
@@ -39,13 +39,6 @@ def add_subcommmand(subparsers):
         metavar="CHAR",
         help="Define the field separator (default: \\t).",
     )
-    parser.add_argument(
-        "--all",
-        "-a",
-        default=False,
-        action="store_true",
-        help="Do not filter duplicate entries.",
-    )
```
dlaehnemann, line 48: Was this removal intentional?

dlaehnemann (approved):
Looks good to me. One question in the code, and:

* Does this need a test, or is the respective functionality covered by the other commands, anyway?

vembrane/modules/table.py
tedil: Yes, I was confused that the short options clashed (`-a`), then had a look where `--all` was used. Turns out: nowhere.

PR #143, opened 2023-03-31
Address some mypy errors and hints

tedil:
This PR addresses those mypy errors and hints that are easy to fix without overcomplicating the code too much.


vembrane/common.py
```diff
@@ -35,6 +35,7 @@ def single_outer(s: str) -> bool:
         return True
     elif "'" in s:
         return False
+    return True
```
dlaehnemann, line 38: Does this addition make sense?

dlaehnemann (approved):
That's a lot of types. I have one question, but no objections.

PR #146, opened 2023-04-27
fix: When a record is missing an entry for ANN, replace with ||||…

tedil:
When
- the VCF header defines the `ANN` `INFO` field,
- the expression makes use of `ANN` and
- a record does not have any value defined for `INFO["ANN"]`
this resulted in an error.
With this change (only for the case described above) `ANN` is set to `"|" * n` where `n` is the number of `ANN` keys defined in the header and a warning is issued.


FelixMoelder (approved):
This appears valid to me. Testing worked well and a short warning is also helpful.

PR #153, opened 2023-07-24
chore!: Replace isort and flake8 with ruff, apply ruff lints, update dependencies, bump min python version to 3.10

tedil:


vembrane/modules/annotate.py
```diff
@@ -67,8 +70,8 @@ def annotate_vcf(
 
     available_chromsomes = set(np.unique(ann_data[config_chrom_column]))
 
-    tree = dict()
-    chrom_ann_data = dict()
+    tree = {}
+    chrom_ann_data = {}
```
dlaehnemann, line 74: I guess this was an automatic change by ruff, right? Just thinking that `dict()` is clear, while `{}` can theoretically be syntax for a `dict` or a `set`, right?

vembrane/modules/tag.py
```diff
@@ -1,8 +1,9 @@
 import re
 import sys
+from collections.abc import Iterator
 from itertools import chain, islice
 from sys import stderr
-from typing import Dict, Iterator, Set, Tuple
+from types import MappingProxyType
```
dlaehnemann, line 6: Just curious: What are these `MappingProxyType`s?

dlaehnemann (approved):
No objections here, this rather looks like mostly syntax updates, right?

Just two questions, but mostly as learning opportunities for me... :D

vembrane/modules/annotate.py
tedil: I think `{}` is always a dict, but I'm not sure. In any case, we may just as well add the corresponding type annotations, i.e. `tree: dict[foo, bar] = {}`.

vembrane/modules/tag.py
tedil: Basically a read-only dict. In our case, these are mainly used as wrappers of empty dicts used as default keyword args (which supply additional, fixed information, such as `auxiliary`). For example, when you invoke `filter(..., auxiliary=foo)` with `foo = {}`, you could mutate `foo` inside of `filter`, potentially causing unwanted behaviour/side-effects etc. Therefore, we wrap that empty dict in a `MappingProxyType`, so no new keys etc can be inserted.

vembrane/modules/annotate.py
dlaehnemann: Yeah, `{}` is always an empty `dict()`, but once you put something in there, it can be a set, e.g. `{'a'}`. So the syntax is clearly defined, I was just thinking it can be confusing to human code readers. So type annotations probably don't hurt. But also, this is a very minor concern...

tedil: I added type annotations here (and also at some other places), so that it's clear we're dealing with dicts.

tedil:
The issue with py3.11 + macOS https://github.com/pysam-developers/pysam/issues/1205  has been fixed in https://github.com/pysam-developers/pysam/pull/1207
Now all we need to do is wait for a new release ;)

Issue #154, opened 2023-07-28
please clarify semantics or add check?


ptrebert:
Hi,
I just ran into the issue that this filter
```
vembrane filter '(is_hom_ref(s) for s in SAMPLES)' input.vcf
````
returned calls that were actually not hom ref calls. It took me a few seconds to realize that I was missing the `all()` here to aggregate the created iterable of bools into a single bool. Maybe there is a way to catch that situation and throw an error?

Without knowing the internals, it looks a bit like this happened:
```
if [False]:
    # evaluating to True because non-empty list
```

Best,
Peter

**Comment by tedil** [comment]
Hi Peter!

You correctly noticed that it didn't return the expected results because the bools generated in your expression weren't aggregated into a single bool (using e.g. `all` or `any`). 
The reason for this lies in Python's concept of truthy and falsy values, which can be found in the Python docs under [Truth Value Testing](https://docs.python.org/3/library/stdtypes.html#truth-value-testing).
In Python, any object is by default considered "truthy", e.g. `bool(some_object) is True`, as long as `some_object` does not define its own `__bool__` or `__len__` implementation.
Consequently, when your filter expression resulted in a [generator](https://docs.python.org/3/glossary.html#term-generator-expression) of booleans, it was considered truthy, leading to the unexpected behaviour you encountered.
```python
assert bool(True for _ in range(4)) is True      # because a generator object is truthy 
assert bool(False for _ in range(4)) is True     # same as above
assert bool(False for _ in range(0)) is True     # same as above
```
However, using a [list comprehension](https://docs.python.org/3/glossary.html#term-list-comprehension) instead, you'd get the following behaviour:
```python
assert bool([True for _ in range(4)]) is True    # because the list is not empty
assert bool([False for _ in range(4)]) is True   # same as above
assert bool([False for _ in range(0)]) is False  # because the list is empty
```

You suggested catching this scenario and possibly throwing an error. While that might seem like a good idea, it's not as straightforward as it appears. The filter expressions are entirely user-defined and can be any arbitrary, valid Python expression.
As a result, we can't make assumptions about how users intend to handle the results. There could be valid use cases where writing a filter making use of truthy values is precisely what's needed, and enforcing an error in such situations (which are not easy to detect anyways!) would be limiting.

I hope that wasn't too confusingly worded and helped you a little; if you have any more questions, just ask!
(Also let me know if I can close this issue)

ptrebert:
Hi Till,
[...]

PR #155, opened 2023-07-28
Enhancement: Raise error for filter expressions which do not evaluate to `bool`

tedil:
This PR addresses #154 by adding a check on the result of the filter expression. If it's `bool`, all is well, otherwise, print a warning but continue, because someone might want to use a truth-y value as a filter.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,7 @@ options:
 
 ### Filter expression
 The filter expression can be any valid python expression that evaluates to a value of *type* `bool`.
+If you want to use truthy values, you need to wrap the expression in `bool()`, or aggregate multiple values via `any()` or `all()`.
 
 However, functions and symbols available have been restricted to the following:
 
--- a/README.md
+++ b/README.md
@@ -42,7 +42,9 @@ options:
 
 
 ### Filter expression
-The filter expression can be any valid python expression that evaluates to `bool`. However, functions and symbols available have been restricted to the following:
+The filter expression can be any valid python expression that evaluates to a value of *type* `bool`.
+
+However, functions and symbols available have been restricted to the following:
 
  * `all`, `any`
  * `abs`, `len`, `max`, `min`, `round`, `sum`
--- a/vembrane/errors.py
+++ b/vembrane/errors.py
@@ -1,3 +1,5 @@
+from typing import Any
+
 from pysam import VariantRecord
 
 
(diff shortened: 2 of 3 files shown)


tedil:
As an alternative, we could also simply force booleans via `keep = bool(self._func())`.

dlaehnemann:
I think a warning is a good idea, here. It should make it easier for users to catch this as unwanted behaviour and allow to use this behaviour if users need it.

christopher-schroeder:
I agree. Maybe an option to disable specific warnings? That should satisfy everyone

dlaehnemann:
And maybe also an addition to the `README.md` docs along the lines that @ptrebert suggested:
https://github.com/vembrane/vembrane/issues/154#issuecomment-1655621757

I would link out to the Truth Value Testing docs of Python that @tedil mentioned and inlcude the statement that a non-empty list will evaluate to `True`.

tedil:
I also thought about whether it is possible to derive from the expression itself whether its result is of type bool or not, but since Python is not a statically typed language, it's probably not particularly easy to get right. 
Instead, one could peek at the first record, apply the filter expression and check that result (and then issue a warning/exit early, whatever).
Though I can imagine that, potentially for some VCF file and expression combinations, the expression may result in different types for different records (such as `0 if INFO['foo'] else "bar"`, which does not make much sense, though.).

Perhaps raising an error instead of a warning would be indeed better, as Peter suggested.
If anyone really wants to rely on truthy values, wrapping the expression in `bool(…)` is always possible and way more explicit.

README.md
```diff
@@ -42,7 +42,9 @@ options:
 
 
 ### Filter expression
-The filter expression can be any valid python expression that evaluates to `bool`. However, functions and symbols available have been restricted to the following:
+The filter expression can be any valid python expression that evaluates to a value of *type* `bool`.
```
dlaehnemann, line 45:
```suggestion
The filter expression can be any valid python expression that evaluates to a value of *type* `bool`.
If you want to use truthy values, you need to wrap the expression in `bool()`, or aggregate multiple values via `any()` or `all()`.
```

dlaehnemann (approved):
I like this approach, it seems like the most sensible. Two things:

1. I would include the new `NonBoolTypeError` help message's info in the main docs as well, as I think this is helpful information for any user and might even make vembrane's behaviour clearer.
2. We should probably bump the major version, as this changes how certain expressions are handled.

dlaehnemann:
That almost sounds like a new Application Note, right? :D

PR #160, opened 2023-09-14
Multi backend

christopher-schroeder:
During the new cyvcf2 branch, Till and I recognize, that it is probably better to let the user decide the backend.
This draft here introduces an abstraction layer and with it several pysam wrapper to be able to create backends with different libraries and address them by polymorphism.

Its not complete at all, currently only

```
vembrane filter "True" exmaple.vcf
```

work as many functions are not implemented yet.


vembrane/common.py
```diff
@@ -42,27 +43,25 @@ def normalize(s: str) -> str:
...
-                    str.strip,
-                    rec.get("Description").strip('"').split(separator)[1].split("|"),
-                )
+    if header.contains_generic("VEP"):
+        separator = ":"
+    if h := header.infos.get(ann_key):
+        return list(
+            map(
+                str.strip,
+                h.get("Description").strip('"').split(separator)[1].split("|"),
```
tedil, line 54: I always felt like this method seems a bit brittle, but it works, so it's fine ;D
However, we do not check anywhere for multiple different annotation keys, e.g. if snpeff or something else adds a CSQ or EFF annotation entry etc.

vembrane/backend/backend_pysam.py
```diff
@@ -0,0 +1,273 @@
...
+        # record.stop is pysams unified approach to get the end position of a variant.
+        # It either considers the alt allele or the END field, depending on the record.
+        # Stop is 0-based, but for consistency with POS we convert into 1-based.
+        # Since for 1-based coordinates, the expectation is that END is inclusive
+        # instead of exclusive (as it is with 0-based), we do not need to add 1.
+        return record.stop
+
+
+def is_bnd_record(record: VariantRecord) -> bool:
+    return "SVTYPE" in record.info and record.info.get("SVTYPE", None) == "BND"
```
tedil, line 273: This is a code duplication (see method of same name in `common.py`) and can probably be removed.

vembrane/errors.py
```diff
@@ -13,7 +13,7 @@ def __str__(self):
 class UnknownAnnotation(VembraneError):
     """Unknown annotation entry"""
 
-    def __init__(self, record_idx: int, record: VariantRecord, key: str):
+    def __init__(self, record_idx: int, record, key: str):
```
tedil, line 16: This (and all other methods below) are missing the type annotation for `record` (which I guess has to do with the import at line 3 is commented out?)

vembrane/globals.py
```diff
@@ -16,6 +16,7 @@
         abs,
         all,
         any,
+        type,
```
tedil, line 19:
```suggestion
```

vembrane/modules/annotate.py
```diff
@@ -41,6 +41,14 @@ def add_subcommmand(subparsers):
         choices=["vcf", "bcf", "uncompressed-bcf"],
         help="Output format.",
     )
+    parser.add_argument(
+        "--backend",
+        "-b",
+        default="pysam",
```
tedil, line 47: As mentioned above, probably use cyvcf2 as the default.

vembrane/modules/annotate.py
```diff
@@ -41,6 +41,14 @@ def add_subcommmand(subparsers):
...
+        type=Backend.from_string,
+        choices=[Backend.pysam, Backend.cyvcf2],
```
tedil, line 49: ... and then also swap the order of these two (which also happens to be alphabetical then)

vembrane/modules/filter.py
```diff
@@ -116,15 +119,23 @@ def add_subcommmand(subparsers):
...
         "Example: `--overwrite-number-format DP=2`",
     )
+    parser.add_argument(
+        "--backend",
+        "-b",
+        default="pysam",
+        type=Backend.from_string,
+        choices=[Backend.pysam, Backend.cyvcf2],
+        help="Set the backend library.",
+    )
```
tedil, line 129: I'd love to get rid of the code duplication in the different modules somehow.

vembrane/representations.py
```diff
@@ -44,116 +27,117 @@ def get(self, item, default=NA):
...
+#                         raise UnknownInfoField(self._record_idx, self._record, item)
+#                 else:
+#                     value = type_info(
+#                         untyped_value,
+#                         self._header_info_fields[item],
+#                         item,
+#                         self._record_idx,
+#                     )
+#             self._info_dict[item] = value
+#             return value
```
tedil, line 134: This can be removed.
```suggestion
```

vembrane/representations.py
```diff
@@ -44,116 +27,117 @@ def get(self, item, default=NA):
...
+#         record_idx: int,
+#         record: VCFRecord,
+#         header_format_fields: Dict[str, str],
+#     ):
+#         self._record = record
+#         self._record_idx = record_idx
+#         self._header_format_fields = header_format_fields
+#         self._record_format = record.format
+
+#     @cache
```
tedil, line 74: The `__getitem__` method has a `@cache` decorator, not sure if that was actually beneficial or not and if we should keep in the new backend specific formats classes?

vembrane/representations.py
```diff
@@ -270,33 +253,30 @@ def __init__(
...
-            for kind in set(r.type for r in header.records)
-        }
+        # self._numbers = {
+        #     category: {
+        #         record_name: overwrite_number.get(category, {}).get(record_name)
+        #         or record.get("Number")
+        #         for record_name, record in category_records.items()
+        #     }
+        #     for category, category_records in header.records.items()
+        # }
```
tedil, line 263: This is now handled elsewhere and can be removed, correct?

vembrane/representations.py
```diff
@@ -355,22 +335,24 @@ def _get_filter(self) -> List[str]:
...
-            self.idx,
-            self.record,
-            self._header_format_fields,
-        )
+    def _get_format(self) -> VCFRecordFormats:
+        # value = Formats(
+        #     self.idx,
+        #     self.record,
+        #     self._header_format_fields,
+        # )
```
tedil, line 354:
```suggestion
```

tedil (changes requested):
We're almost done! Thanks a lot @christopher-schroeder :)
There are some parts which need small changes, but most of the changes look fine.
What I think is missing are some more tests for:

- making sure record.stop is identical between pysam and cyvcf2 backends for uncommon ref+alt case combinations, especially for structural variation (perhaps there are good and usable examples in the vcf spec?)
- annotate does not have any tests yet (and is due to change anyways), but at least one test would be great.
- tests for resetting the reader in the bnd/sorted output 2-pass case (both for files and streams, with the stream ones expected to fail)

What we also still need to discuss is whether the NoValueDict class should raise a KeyError or silently return False.

vembrane/backend/backend_cyvcf2.py
```diff
@@ -0,0 +1,278 @@
...
+class Cyvcf2Reader(VCFReader):
+    def __init__(
+        self,
+        filename: str,
+        overwrite_number: Dict[str, Dict[str, str]] = {},
+    ):
+        self.filename = filename
+        self._file = VCF(self.filename)
+        self._header = Cyvcf2Header(self, overwrite_number)
+        self._overwrite_number = overwrite_number
```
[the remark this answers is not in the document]
christopher-schroeder: I added __slots__, but pytest doesn't throw any error, when the field is not in slots. Thats super weird. Do you know how to make pytest include slots?

vembrane/backend/backend_pysam.py
```diff
@@ -0,0 +1,273 @@
...
+        header: VCFHeader,
+    ):
+        self._record = record
+        self._header = header
+
+    def __getitem__(self, key):
+        if key == "END":
+            return get_end(self._record)
+        if key not in self._header.infos.keys():
+            raise UnknownInfoField(0, self._record, key)  # TODO: self._record_idx
```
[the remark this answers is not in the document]
christopher-schroeder: TODO done

vembrane/backend/backend_pysam.py
```diff
@@ -0,0 +1,273 @@
...
+        if not self.__contains__(sample):
+            raise UnknownSample(self._record_idx, self._record, sample)
+        meta = self._header.formats[self._format_key]
+        return type_info(self._record.samples[sample][self._format_key], meta["Number"])
+
+    def __setitem__(self, key, value):
+        self._record.format[key] = value
+
+    def __contains__(self, sample):
+        return sample in self._record.samples.keys()
```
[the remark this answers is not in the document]
christopher-schroeder: I didn't know either. I've looked it up and VCFv4.2 and from 1.4.2

_If genotype information is present, then the same types of data must be present for all samples. First a FORMAT
field is given specifying the data types and order (colon-separated alphanumeric String). This is followed by one field
per sample, with the colon-separated data in this field corresponding to the types specified in the format_

I think it enforces it.

vembrane/representations.py
christopher-schroeder: I added the @cache decorator, because the old code was actually a self implemented cache. I now add the cache decorator where I think it might be helpful for some queries

vembrane/representations.py
```diff
@@ -270,33 +253,30 @@ def __init__(
...
+        #         for record_name, record in category_records.items()
+        #     }
+        #     for category, category_records in header.records.items()
+        # }
 
         # always explicitly set "Number" for certain fields
         # which get special pysam treatment:
         # - `FORMAT["GT"]` is always parsed as a list of integer values
-        self._numbers.get("FORMAT", {})["GT"] = "."
+        # self._numbers.get("FORMAT", {})["GT"] = "."
```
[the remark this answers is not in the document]
christopher-schroeder: now it is

vembrane/backend/backend_cyvcf2.py
```diff
@@ -0,0 +1,278 @@
...
+        meta = self._header.formats[self._format_key]
+        number = meta["Number"]
+        if meta["Type"] == "String":
+            value.item()
+            if not number == "1":
+                value = value.split(",")
+        if number == "1":
+            if isinstance(value, np.ndarray):
+                value = value[0].item()
+            # cyvcf2 gives min int for unknown integer values
```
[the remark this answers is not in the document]
christopher-schroeder: Its not :)

vembrane/backend/backend_cyvcf2.py
```diff
@@ -0,0 +1,278 @@
...
+        header: Cyvcf2Header,
+    ):
+        self._record = record
+        self._header = header
+
+    def __getitem__(self, key):
+        # if key == "END":
+        #     return get_end(self._record)
+        if key not in self._header.infos.keys():
+            raise UnknownInfoField(0, self._record, key)  # TODO: self._record_idx
```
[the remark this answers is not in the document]
christopher-schroeder: record idx now implemented, together with code simplification

vembrane/backend/backend_cyvcf2.py
```diff
@@ -0,0 +1,278 @@
...
+
+        return type_info(value, number)
+        # if number == "A" and not isinstance(value, list):
+        #     value = tuple([value])
+
+    def __setitem__(self, key, value):
+        # for some reasons cyvcf2 doesn't split String lists, a known circumstance
+        meta = self._header.infos[key]
+        number, typ = meta["Number"], meta["Type"]
+        if typ == "String" and number == ".":
```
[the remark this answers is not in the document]
christopher-schroeder: done

vembrane/backend/backend_pysam.py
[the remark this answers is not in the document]
christopher-schroeder: the implementation is now a OrderedSet (OrderedDict). That kind of caches it.

vembrane/backend/backend_pysam.py
```diff
@@ -0,0 +1,273 @@
...
+    def samples(self):
+        return self._header.samples
+
+    @property
+    def filters(self):
+        return self._header.filters
+
+    @property
+    def records(self):
+        return self._data_category
```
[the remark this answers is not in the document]
christopher-schroeder: I think the term `_header_records is even worse.

vembrane/backend/base.py
```diff
@@ -0,0 +1,272 @@
...
+
+    def __repr__(self):
+        return str(self)
+
+    @staticmethod
+    def from_string(s):
+        try:
+            return Backend[s]
+        except KeyError:
+            return s
```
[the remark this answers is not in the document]
christopher-schroeder: No this is a hack to make argparse properly let you choose from enums, that are adressed by string. I don't really know whats going on here, but if you raise an error instead of s, argparse doesn't print the correct information anymore in case the user doesn't choose correctly.

vembrane/common.py
christopher-schroeder: Unfortunately there are also other tools that annotate in a similar way. I would like to move the discussion about that to the follow up PR, where I introduce a way to use user defined custom annotations by an external provided yaml.

vembrane/backend/backend_pysam.py
christopher-schroeder: Yes it it. I am avoiding a circular import here. The only solution I can come up with is splitting the common file into multiple files. Any ideas are welcome!

vembrane/common.py
```diff
@@ -133,3 +132,23 @@ def read_set(path: str) -> Set[str]:
...
+
+def create_writer(
+    filename: str, fmt: str, template: VCFReader, backend: Backend = Backend.pysam
+):
+    if backend == Backend.pysam:
+        return PysamWriter(filename, fmt, template)
+    elif backend == Backend.cyvcf2:
+        return Cyvcf2Writer(filename, fmt, template)
+    else:
+        raise ValueError(f"{backend} is not a known backend.")
```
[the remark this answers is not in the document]
christopher-schroeder: I agree.

vembrane/errors.py
christopher-schroeder: again circular import, I don't know how to handle

vembrane/globals.py
christopher-schroeder: sorry that was for debugging

vembrane/backend/backend_pysam.py
tedil: Oh, right. Hm, I guess we can just leave it as is for now

vembrane/backend/base.py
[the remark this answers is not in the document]
tedil: Ah, I see, I didn't know that!

vembrane/common.py
tedil: Sure, that makes sense

vembrane/backend/backend_pysam.py
christopher-schroeder: I just move the functions to VCFRecord

vembrane/modules/annotate.py
christopher-schroeder: cyvcf2 is default now

vembrane/modules/annotate.py
christopher-schroeder: cyvcf2 is default now and order correct

vembrane/representations.py
christopher-schroeder: yes

vembrane/modules/filter.py
christopher-schroeder: I moved this code to a common function in common.py

tests/testcases/filter/test_type_and_number_combinations/config.yaml
```diff
@@ -0,0 +1,36 @@
+function: "filter"
```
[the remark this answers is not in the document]
christopher-schroeder: Done

vembrane/backend/backend_cyvcf2.py
```diff
@@ -0,0 +1,278 @@
...
+    # @property
+    # def header(self):
+    #     return self._header
+
+    def reset(self):
+        # cyvcv2 doesnt have a reset function
+        self._file.close()
+        self._file = VCF(self.filename)
+        self._header = Cyvcf2Header(self, self._overwrite_number)
+        # TODO: may this workaround lead to problems?
```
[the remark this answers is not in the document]
christopher-schroeder: I put this to the Roadmap

vembrane/backend/backend_cyvcf2.py
```diff
@@ -0,0 +1,278 @@
...
+
+    @property
+    def position(self) -> int:
+        return self._record.POS
+
+    @property
+    def stop(self) -> int:
+        if "END" in self.info:
+            return self.info["END"]
+        return self._record.POS + len(self._record.REF) - 1
```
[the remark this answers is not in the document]
christopher-schroeder: It is. Tested with a Manta called vcf

vembrane/backend/backend_pysam.py
```diff
@@ -0,0 +1,273 @@
...
+    def contig(self) -> str:
+        return self._record.chrom
+
+    @property
+    def position(self) -> int:
+        return self._record.pos
+
+    @property
+    def stop(self) -> int:
+        return self._record.stop
```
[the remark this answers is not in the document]
christopher-schroeder: Added a testcase for END

vembrane/backend/backend_cyvcf2.py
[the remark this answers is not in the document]
christopher-schroeder: nevermind, it seems to work now

vembrane/backend/backend_cyvcf2.py
```diff
@@ -0,0 +1,278 @@
...
+
+    @property
+    def samples(self):
+        return self._file.samples
+
+    def __repr__(self):
+        return self._record.__repr__()
+
+    def __str__(self):
+        return self._record.__str__()
```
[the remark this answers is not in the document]
christopher-schroeder: I moved the code to base. By now this should be identical for pysam and cyvcf2 as it is just the vcf representation of the record for both backends

vembrane/backend/backend_pysam.py
```diff
@@ -0,0 +1,273 @@
...
+
+    @property
+    def samples(self):
+        return self._record.samples
+
+    def __repr__(self):
+        return self._record.__repr__()
+
+    def __str__(self):
+        return self._record.__str__()
```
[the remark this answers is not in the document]
christopher-schroeder: dito

vembrane/backend/base.py
```diff
@@ -0,0 +1,272 @@
...
+    pass
+
+
+class NoValueDict:
+    def __contains__(self, item):
+        try:
+            value = self[item]
+        except KeyError:
+            return False
+        return value is not NA
```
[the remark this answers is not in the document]
christopher-schroeder: So the consense was `false` and Sven a warning, which itself is not implemented. Do we want to have such a warning for each missing value in each read?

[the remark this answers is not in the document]
tedil: If a VCF file has `some_random_key` specified in its header, every record should have that field set, which is why the VCF spec has that special value MISSING (i.e. `.`).
If one of a million records for some reason does not have `some_random_key` specified, I'd just take it to mean `some_random_key=.`, which means that the information is missing for that key.

I have added a warning printing both the record id and the record itself in that case; that might be annoying for files which have a lot of records which simply omit info keys, instead of using `.` as corresponding values.

I have tried finding information about this in the VCF spec, but couldn't find anything definitive as to what is the preferred way. `bcftools` at least does not complain about this case.

PR #176, opened 2024-09-19
fix: properly handle MATEID presented as single-item

johanneskoester:
This is relevant for the SAVANA caller, which encodes MATEID as follows:
`##INFO=<ID=MATEID,Number=1,Type=String,Description="ID of mate breakends">`


FelixMoelder (approved):

tedil:
This should have targeted the dev-1.0 branch, as the main branch currently contains the changes for vembrane 2.0.
We can do a rebase and then release that I guess

PR #184, opened 2025-02-21
feat: new subcommand for generating structured output from vcf using a yte template

johanneskoester:


tedil:
Uh, super nice!

tedil (changes requested):
This is very nice, thank you!
The only things I believe are missing at the moment are:
 - updating the readme to include the new subcommand + some examples
 - tests which show slightly more complex expressions

johanneskoester:
I've added a section to the README and a more complex test. This showed me that I had so far not considered ANN in the code. However, it seems like the ANN field is not found now from within YTE, although I tried to follow the same logic as in table and filter.

johanneskoester:
I had to raise the minimum python version to 3.12, because that was the only way to pass both vembranes Envrionment as well as yte's intermediate variables to `exec` and `eval` without running into "undefined name" errors.

johanneskoester:
No clue why exactly that was the case to be honest, I did not anticipate a difference with that in the release of Python 3.12

tedil (approved):
LGTM! Thanks for this functionality.

PR #185, opened 2025-02-22
build: Switch to `uv` with `hatchling` as the build backend

tedil:


johanneskoester (commented):
LGTM, but, out of interest, why not just directly use pixi? It handles pyproject.toml as well. Further, it has workflow capabilities which could be used to get the same as with tox I guess.

tedil:
Does pixi already use uv? Not sure how up to date this blog entry is: https://prefix.dev/blog/uv_in_pixi

PR #196, opened 2025-06-13
docs: fix examples with missing `SAMPLES[0]` reference for FORMAT fields

dlaehnemann:


tedil:
Hm, these example should work, the indexing should be fine with both an index (`0`) and a sample name (`SAMPLES[0]`). If it isn't, this is a bug/regression, not a documentation issue.

dlaehnemann:
OK, let's look into creating a test case for this, then.

PR #197, opened 2025-06-24
feat: new subcommand for sorting VCF files via one or multiple Python expressions

johanneskoester:
The `sort` subcommand allows to sort VCF/BCF files via keys defined as Python expressions.
The Python expressions are analogous to expressions in other vembrane commands.
This feature loads the entire VCF file into memory in order to maximize performance.
It is thus meant to sort small, already filtered VCF files, e.g. for **prioritizing records for the human eye**.
For large VCF files, the only relevant sorting is usually by position, which is better done with e.g. bcftools (and usually the sorting that variant callers output).

### Example

The following command sorts by gnomad_AF (ascending) and REVEL score (descending).
Therreby the gnomad allele frequencies are quantized in order to group them into 10 bins.

```bash
vembrane sort calls.vcf "round(ANN['gnomAD_AF'], 1), -ANN['REVEL']"  > prioritized.vcf
```

### TODO
* [x] add testcase(s)
* [x] handle NAs (most convenient is to sort them to the end


tests/test_vembrane.py
```diff
@@ -99,8 +101,18 @@ def test_command(testcase: os.PathLike, backend: Backend):
...
+                or args.command == "tag"
+                or args.command == "sort"
+            ):
+                match args.command:
+                    case "filter":
+                        module = filter
+                    case "tag":
+                        module = tag
+                    case "sort":
+                        module = sort
```
tedil, line 115: unrelated, but perhaps we can use a dictionary here instead?
```py
commands = {
  "filter": filter,
  …
}
module = commands[args.command]
…
```
and move the handling of the different module groups (filter/tag/sort vs table vs structured/fhir) to their own functions.

README.md
```diff
@@ -356,9 +356,60 @@ Exemplary invocation: `vembrane annotate example.yaml example.bcf > annotated.vc
...
+```
+usage: vembrane sort [-h] [--output OUTPUT] [--output-fmt {vcf,bcf,uncompressed-bcf}] [--annotation-key FIELDNAME] [--aux NAME=PATH] [--ontology PATH]
+                     [--overwrite-number-info FIELD=NUMBER] [--overwrite-number-format FIELD=NUMBER] [--backend {cyvcf2,pysam}]
+                     [vcf] sort_key [sort_key ...]
+
+positional arguments:
+  vcf                   The VCF/BCF file containing the variants. If not specified, reads from STDIN.
+  sort_key              Python expressions returning orderable values to sort the VCF records by (ascending, smallest values coming first).
+                        If multiple expressions are provided, they are prioritized from left to right with lowest priority on the right.
+                        NA/NaN values are sorted to the end.
```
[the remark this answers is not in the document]
johanneskoester: Good point. Maybe better to rather always require the filepath or - to be specified.

tests/test_vembrane.py
johanneskoester: I agree. But I guess rather in a separate PR?

README.md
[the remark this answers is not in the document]
johanneskoester: Further thought about the sort_key. Maybe it is indeed better to model it in the same way as the column expressions for the table subcommand. In the end, one has to type less letters, and since we will support the subcommand in a wrapper for Snakemake, that one will also be easy to use for people with sort keys provided as list given to params.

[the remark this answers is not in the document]
johanneskoester:
```
cat example.vcf | vembrane sort '-POS, len(REF)' 
```
And this will also always work, regardless of whether the user omits - or has it.

README.md
```diff
@@ -356,9 +356,60 @@ Exemplary invocation: `vembrane annotate example.yaml example.bcf > annotated.vc
...
+### Usage
+
+```
+usage: vembrane sort [-h] [--output OUTPUT] [--output-fmt {vcf,bcf,uncompressed-bcf}] [--annotation-key FIELDNAME] [--aux NAME=PATH] [--ontology PATH]
+                     [--overwrite-number-info FIELD=NUMBER] [--overwrite-number-format FIELD=NUMBER] [--backend {cyvcf2,pysam}]
+                     [vcf] sort_key [sort_key ...]
+
+positional arguments:
+  vcf                   The VCF/BCF file containing the variants. If not specified, reads from STDIN.
+  sort_key              Python expressions returning orderable values to sort the VCF records by (ascending, smallest values coming first).
```
tedil, line 377: expressions → expression
+ mention tuple style / comma separation?

vembrane/modules/sort.py
```diff
@@ -0,0 +1,138 @@
...
+                ),
+            )
+
+            annotation = Annotation(args.annotation_key, reader.header)
+
+            # load records into memory
+            # since sorting by something else than the position is only reasonable
+            # for already small, filtered VCF files, this is not an issue in practice.
+            # TODO: auto-detect larger files and use a mergesort library for those?
+            records = list(enumerate(reader))
```
tedil, line 90: Turn this into an issue so we don't lose track. Right now, most (all?) vembrane commands support streaming, and this being an exception is somewhat unexpected for the end-user (who may or may not read the disclaimer in the documentation / help)

vembrane/modules/sort.py
```diff
@@ -0,0 +1,138 @@
...
+            if sort_keys.expression_annotations():
+                # if the expression refers to annotations, we obtain the
+                # minimal keys across all annotations
+                annotations = annotation.get_record_annotations(idx, record)
+
+                def eval_with_ann(annotation):
+                    sort_keys.update_annotation(annotation)
+                    return handle_na(eval(sort_keys.compiled, sort_keys))
+
+                return min(map(eval_with_ann, annotations))
```
tedil, line 122: So annotations aren't sorted within the list of annotations, just the records based on the "smallest" ann? Is that mentioned / explained somewhere? How much work would it be to do that? Because at least `filter` also operates on that level (filtering out individual annotations within a single record)

vembrane/modules/structured.py
```diff
@@ -16,15 +16,19 @@
 
 
 def add_subcommand(subparsers):
-    parser = subparsers.add_parser("structured")
+    parser = subparsers.add_parser(
+        "structured",
+        description="Converts VCF records into structured data in YAML, JSON or JSONL "
+        "format given a YAML/YTE template.",
+    )
```
tedil, line 23: out of scope here, but glad to have this ;D

README.md
```diff
@@ -356,9 +356,60 @@ Exemplary invocation: `vembrane annotate example.yaml example.bcf > annotated.vc
...
+                        Set the backend library.
+```
+
+### Example
+
+The following command sorts by gnomad_AF (ascending) and REVEL score (descending).
+Thereby the gnomad allele frequencies are quantized in order to group them into 10 bins.
+
+```bash
+vembrane sort calls.vcf "round(ANN['gnomad_AF'], 1)" "-ANN['REVEL']" > prioritized.vcf
```
tedil, line 407:
```suggestion
vembrane sort calls.vcf 'round(ANN["gnomad_AF"], 1), -ANN["REVEL"]' > prioritized.vcf
```

README.md
```diff
@@ -356,9 +356,60 @@ Exemplary invocation: `vembrane annotate example.yaml example.bcf > annotated.vc
...
+  --overwrite-number-format FIELD=NUMBER
+                        Overwrite the number specification for FORMAT fields given in the VCF header. Example: `--overwrite-number-format DP=2`
+  --backend {cyvcf2,pysam}, -b {cyvcf2,pysam}
+                        Set the backend library.
+```
+
+### Example
+
+The following command sorts by gnomad_AF (ascending) and REVEL score (descending).
+Thereby the gnomad allele frequencies are quantized in order to group them into 10 bins.
```
tedil, line 404: I tried making it clear that it's _records_ that get sorted, not individual values or something else. Feel free to adapt as you like
```suggestion
The following command sorts records first by `gnomad_AF` (binned and ascending), and then by `REVEL` score (descending).
The descending sort is achieved by negating the `REVEL` value (`-ANN['REVEL']`) in the key expression.
```

README.md
```diff
@@ -356,9 +356,60 @@ Exemplary invocation: `vembrane annotate example.yaml example.bcf > annotated.vc
...
 
+
+## `vembrane sort`
+
+The `sort` subcommand allows to sort VCF/BCF files via keys defined as Python expressions.
+The Python expressions are analogous to expressions in other vembrane commands.
+This feature loads the entire VCF file into memory in order to maximize performance.
+It is thus meant to sort small, already filtered VCF files, e.g. for **prioritizing records for the human eye**.
+For large VCF files, the only relevant sorting is usually by position, which is better done with e.g. bcftools (and usually the sorting that variant callers output).
+
```
tedil, line 367: Mention the default sort order (ascending?)

README.md
tedil: Also, how do I invert (i.e. toggle between ascending/descending) non number types? I assume `-"string"` and `-FALSE` will not work?
(→ also add an example for these cases)

README.md
```diff
@@ -356,9 +356,60 @@ Exemplary invocation: `vembrane annotate example.yaml example.bcf > annotated.vc
...
+                     [--overwrite-number-info FIELD=NUMBER] [--overwrite-number-format FIELD=NUMBER] [--backend {cyvcf2,pysam}]
+                     [vcf] sort_key [sort_key ...]
+
+positional arguments:
+  vcf                   The VCF/BCF file containing the variants. If not specified, reads from STDIN.
+  sort_key              Python expressions returning orderable values to sort the VCF records by (ascending, smallest values coming first).
+                        If multiple expressions are provided, they are prioritized from left to right with lowest priority on the right.
+                        NA/NaN values are sorted to the end.
+
+options:
```
tedil, line 381: Do we want to include an option for choosing between ascending / descending?

tests/testcases/sort/test_01/test.vcf
```diff
@@ -0,0 +1,245 @@
...
+##INFO=<ID=known_CLIN_not_provided,Number=0,Type=Flag,Description="not provided.https://www.ensembl.org/info/genome/variation/phenotype/phenotype_annotation.html#clin_significance">
+##INFO=<ID=dgiDB_drugs,Number=.,Type=String,Description="Combination of gene, drug, interaction types extracted from dgiDB. Each combination is pipe-seperated annotated as GENE|DRUG|TYPE">
+#CHROM	POS	ID	REF	ALT	QUAL	FILTER	INFO	FORMAT	tumor
+1	26696985	rs1283126655	C	T	30	PASS	SVLEN=.;PROB_ARTIFACT=8.16365;PROB_FFPE_ARTIFACT=inf;PROB_PRESENT=1.15636;PROB_ABSENT=10.908;ANN=T|synonymous_variant|LOW|ARID1A|ENSG00000117713|Transcript|ENST00000324856|protein_coding|1/20||ENST00000324856.13:c.582C>T|ENSP00000320485.7:p.Pro194%3D|971|582|194|P|ccC/ccT|rs1283126655||1||SNV|HGNC|HGNC:11110|YES|NM_006015.6|1|P3|CCDS285.1|ENSP00000320485|O14497||UPI0000167B91|1|||PANTHER:PTHR12656&PANTHER:PTHR12656:SF12&MobiDB_lite:mobidb-lite|||||||||||0|0|0|0|0|0|0|0|0|0|gnomAD_AFR&gnomAD_AMR&gnomAD_ASJ&gnomAD_EAS&gnomAD_FIN&gnomAD_NFE&gnomAD_OTH&gnomAD_SAS|||||||||0.0215|likely_pathogenic|0.9974	GT	./.
```
tedil, line 244: I'd like to see testcases with multiple annotations per record, so we can check the "minimal key across annotations" choice thing.

tedil: Regarding this specific one which uses the 'Consequence': I assume these are treated as strings (because they also inherit from string, and that provides the necessary ordering related _functions).
Could we add an example (also to the docs) that allows to sort by an order of strings specified through `--aux` (though that may be a set, not a list… not sure atm)? You can probably ignore this comment

tedil (changes requested):

vembrane/modules/sort.py
johanneskoester: Yes, I think there is no use case for sorting annotations. The purpose is to get the records in the desired order. But, when combining this e.g. with table, it would be a bit weird if the annotations then are sorted within each record. I wonder what would be the best strategy here...

tests/testcases/sort/test_01/test.vcf
johanneskoester: For now, I thought that people could compute the appropriate keys for strings within their expression. Might make the expression long of course. But that again might be mitigated by a follow-up idea for all subcommands, which would be the ability to specify a python module that can offer helper functions containing more complex reusable code. E.g. a function like

```python
def consequence_key():
    match ANN["Consequence"]:
        case "missense_variant": 0
        case "synonymous_variant: 1
```
```

File: vembrane/modules/sort.py (hunk 5 above)
Comment (johanneskoester) [reply vembrane/modules/sort.py:122]: At the moment, I even favor the splitting strategy over the min strategy.

File: vembrane/modules/sort.py [hunk 12]
```diff
@@ -0,0 +1,138 @@
...
+                    or value is None
+                    or (isinstance(value, float) and math.isnan(value))
+                ):
+                    return NAToEnd()
+                return value
+
+            if sort_keys.expression_annotations():
+                # if the expression refers to annotations, we obtain the
+                # minimal keys across all annotations
+                annotations = annotation.get_record_annotations(idx, record)
```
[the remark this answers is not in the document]
johanneskoester: see above

vembrane/modules/sort.py
johanneskoester: So, everything as discussed now. The code automatically uses external sorting in case the VCF exceeds a given size. Also, we have `asc()` and `desc()` decorators for keys that allow to set sort order per tuple item. The default is `asc()` so that it can be omitted for brevity.

johanneskoester: Further, we now have a flag --preserve-annotation-order that disables the default behavior (sorting annotations as well).

johanneskoester: In other words, ready for review @tedil!

docs/sort.md
```diff
@@ -0,0 +1,66 @@
...
+
+```
+usage: vembrane sort [-h] [--output OUTPUT] [--output-fmt {vcf,bcf,uncompressed-bcf}] [--preserve-annotation-order]
+                     [--chunk-size CHUNK_SIZE] [--annotation-key FIELDNAME] [--aux NAME=PATH] [--ontology PATH]
+                     [--overwrite-number-info FIELD=NUMBER] [--overwrite-number-format FIELD=NUMBER]
+                     [--backend {cyvcf2,pysam}]
+                     [vcf] expression
+
+Sort VCF records by one or multiple Python expressions that encode keys for the desired order. This feature loads the entire
+VCF file into memory in order to maximize performance. It is thus meant to sort small, already filtered VCF files, e.g. for
```
tedil, line 19: now that there's the external sort / --chunk-size option, no longer completely true

docs/sort.md
```diff
@@ -0,0 +1,66 @@
...
+                        Output file, if not specified, output is written to STDOUT.
+  --output-fmt {vcf,bcf,uncompressed-bcf}, -O {vcf,bcf,uncompressed-bcf}
+                        Output format.
+  --preserve-annotation-order
+                        If set, annotations are not sorted within the records, but kept in the same order as in the input VCF
+                        file. If not set (default), annotations are sorted within the record according to the given keys if
+                        any of the sort keys given in the python expression refers to an annotation.
+  --chunk-size CHUNK_SIZE
+                        Number of VCF records to sort in memory. If the VCF file exceeds this number of records, external
+                        sorting is used.
```
tedil, line 41: Can argparse include the default here?

docs/sort.md
```diff
@@ -0,0 +1,66 @@
...
+                        Overwrite the number specification for FORMAT fields given in the VCF header. Example: `--overwrite-
+                        number-format DP=2`
+  --backend {cyvcf2,pysam}, -b {cyvcf2,pysam}
+                        Set the backend library.
+```
+
+### Example
+
+The following command sorts records first by `gnomad_AF` (binned and ascending), and then by `REVEL` score (descending).
+The descending sort is achieved by negating the `REVEL` value (`-ANN['REVEL']`) in the key expression.
```
tedil, line 62: Mention that this could also be done with `desc(ANN['REVEL'])` here?

tests/testcases/sort/test_01/config.yaml
```diff
@@ -0,0 +1,4 @@
+function: "sort"
+expression: "ANN['Consequence']"
+chunk-size: "1"
+preserve-annotation-order: true
```
tedil, line 4: I'd like the tests to have more descriptive names!

vembrane/modules/sort.py
```diff
@@ -0,0 +1,359 @@
...
+from vembrane.representations import Annotation, SourceEnvironment
+
+
+def add_subcommand(subparsers):
+    parser = subparsers.add_parser(
+        "sort",
+        help="Sort VCF records by one or multiple Python expressions.",
+        description="Sort VCF records by one or multiple Python expressions that "
+        "encode keys for the desired order. This feature loads the entire VCF file "
+        "into memory in order to maximize performance. It is thus meant to sort small, "
```
tedil, line 30: same as before, now does external sort if num records > chunksize

vembrane/modules/sort.py
```diff
@@ -0,0 +1,359 @@
...
+        action="store_true",
+        help="If set, annotations are not sorted within the records, but kept in the "
+        "same order as in the input VCF file. If not set (default), annotations "
+        "are sorted within the record according to the given keys if any of the sort "
+        "keys given in the python expression refers to an annotation.",
+    )
+    parser.add_argument(
+        "--chunk-size",
+        type=int,
+        default=1000,
```
tedil, line 76: 1000 seems like a small number, I guess a single record only occupies a couple hundred bytes, so should be safe to raise to 10_000 if not even 100_000?

vembrane/modules/sort.py
```diff
@@ -0,0 +1,359 @@
...
+    )
+    parser.add_argument(
+        "expression",
+        type=check_expression,
+        help="Python expression (or tuple of expressions) returning orderable values "
+        "to sort the VCF records by (ascending, smallest values coming first). "
+        "If multiple expressions are provided as a tuple, they are prioritized from "
+        "left to right with lowest priority on the right. "
+        "NA/NaN values are sorted to the end.",
+    )
```
tedil, line 51: these are the wrong way around, expression first, then vcf

vembrane/modules/sort.py
```diff
@@ -0,0 +1,359 @@
...
+        if is_na(other.key):
+            return False
+        return self.key > other.key
+
+    def __gt__(self, other: Self) -> bool:
+        if is_na(self.key):
+            return False
+        if is_na(other.key):
+            return True
+        return self.key < other.key
```
tedil, line 148: does this not mean that NA/None/nan are sorted to the beginning, which is contrary to what the docs state (i.e. NA always at the end)?

tedil (changes requested):

tests/testcases/sort/test_01/config.yaml
johanneskoester: done!

johanneskoester:
I have implemented all suggested changes. Thanks for the thorough assessment. This really made it much better!

tedil (approved):

PR #298, opened 2026-07-08
fix: is_hom_var using 's' instead of 'sample'

tedil:
~TODO: add testcases~


dlaehnemann (approved):
Looks good. And ouch for the underlying bug... 😬

PR #310, opened 2026-07-31
fix: `.get("GT")` returning None and `has_genotype(...)` incorrectly checking for single `NA` and typing issues

tedil:
… and while at it, also fix some abstract base vs implementation method placements and usages of deprecated abstractproperty decorator etc.


dlaehnemann (approved):
The test case and the respective `has_genotype()` changes look valid to me. For the rest, I don't feel qualified.

PR #315, opened 2026-08-19
chore: update pytest, add pytest-sugar to dev group explicitly, update uv.lock

tedil:


dlaehnemann (approved):
What's `pytest-sugar`? Sound sweet.

PR #316, opened 2026-08-19
chore: update ci deps

tedil:


.github/workflows/release-please.yml
```diff
@@ -35,11 +35,11 @@ jobs:
         uses: astral-sh/setup-uv@v7
         with:
           # Install a specific version of uv and enable caching.
-          version: "0.6.2"
+          version: "0.11.30"
           enable-cache: true
 
       - name: Set up Python
-        run: uv python install 3.10  # Build only for Python 3.10
+        run: uv python install 3.12  # Build only for Python 3.12
```
dlaehnemann, line 42: Does it make sense to build for multiple `python` versions with a matrix here, to keep dependency management systems happier? Or does that happen during the pypi-publish step implicitly, anyways?

dlaehnemann (approved):
Looks good, just have a question for a possible improvement...?!

.github/workflows/release-please.yml
tedil: for publishing, we only need a single python version, since uv will simply build an sdist + universal wheel I believe, which aren't really python version dependent.

dlaehnemann: Magic. :sparkles:

PR #326, opened 2026-09-22
fix: properly handle empty VCF/BCF in case of parquet output in vembrane table

johanneskoester:
Before, an empty input VCF/BCF caused a StopIteration exception when calling next().


vembrane/modules/table.py
```diff
@@ -460,7 +460,12 @@ def execute(args):
         elif args.output_fmt == "parquet":
             with smart_open(args.output, "wb") as outfile:
                 chunks = batched(rows, args.parquet_row_group_size)
-                first_chunk = next(chunks)
+                is_empty = False
+                try:
+                    first_chunk = next(chunks)
+                except StopIteration:
+                    is_empty = True
+                    first_chunk = []
```
tedil: is there something similar to a peekable trait/wrapper available in python stdlib? ah, more-itertools has something like this, I believe.

tedil (dismissed):
LGTM, though I'd prefer using more-itertool's `peekable` abstraction I think. Would be a new dependency, I believe

vembrane/modules/table.py
johanneskoester: You are right, this is a very good idea, done.

tedil (approved):

